PR #4, opened 2019-11-28
Update to integrations v4 and added dependency to nria

cristianciutea:
This PR will change the config system to v4 which means that the minimum agent version is 1.8.0. This is required for FF


Makefile-package.mk
```diff
@@ -33,13 +33,11 @@ prep-pkg-env:
...
 	@mkdir -p $(SOURCE_DIR)/var/db/newrelic-infra/newrelic-integrations/bin $(SOURCE_DIR)/etc/newrelic-infra/integrations.d
-	@echo "=== Main === [ prep-pkg-env ]: adding built binaries and configuration and definition files..."
+	@echo "=== Main === [ prep-pkg-env ]: adding built binaries and configuration files..."
 	@cp $(BINS_DIR)/$(BINARY_NAME) $(SOURCE_DIR)/var/db/newrelic-infra/newrelic-integrations/bin
 	@chmod 755 $(SOURCE_DIR)/var/db/newrelic-infra/newrelic-integrations/bin/*
-	@cp ./*-definition.yml $(SOURCE_DIR)/var/db/newrelic-infra/newrelic-integrations/
-	@chmod 644 $(SOURCE_DIR)/var/db/newrelic-infra/newrelic-integrations/*-definition.yml
-	@cp ./*-config.yml.sample $(SOURCE_DIR)/etc/newrelic-infra/integrations.d/
-	@chmod 644 $(SOURCE_DIR)/etc/newrelic-infra/integrations.d/*-config.yml.sample
+	@cp ./*-config.yml $(SOURCE_DIR)/etc/newrelic-infra/integrations.d/
```
davidbrota: Shouldn't we add the -n flag to avoid overwriting the configuration file if it exists?

cristianciutea: I think this is copying the file into the package that we create.
'The fpm `--config-files /etc/newrelic-infra/` option already does the trick 
--config-files CONFIG_FILES   Mark a file in the package as being a config file. This uses 'conffiles' in debs and %config in rpm.
conffiles
One of the most annoying things about software is when you spend a great deal of time and effort customizing a program, only to have an upgrade stomp all over your changes. Debian solves this problem by marking such configuration files as conffiles. [55] When you upgrade a package, you'll be asked whether you want to keep your old configuration files or not.

RPM spec files have a macro, %config, that is used to mark config files so that edits to config files won't get lost during a subsequent upgrade'

davidbrota: 👍

PR #14, opened 2019-12-13
Remove duplicated cpou sampling

varas:
Harmless but redundant operation.


src/nri/sampler.go
```diff
@@ -108,7 +108,6 @@ func (cs *ContainerSampler) SampleAll(i *integration.Integration) error {
 		populate(ms, memory(&metrics.Memory))
 		populate(ms, pids(&metrics.Pids))
 		populate(ms, blkio(&metrics.BlkIO))
-		populate(ms, cpu(&metrics.CPU))
```
varas: already done above https://github.com/newrelic/nri-docker/pull/14/files#diff-8b369eee4576217066ed0ea9bb7c6193L107

PR #19, opened 2020-02-06
Detect cgroup path

cristianciutea:
This pr will add detection of cgroup location:
1. --cgroup_path command line argument
2. checking existence of cgroup usual paths:
	"/sys/fs/cgroup",
	"/cgroup",
3. parsing mounts file

### DIFF
--- a/src/raw/cgroup.go
+++ b/src/raw/cgroup.go
@@ -29,13 +29,13 @@ type cgroupsFetcher struct {
 	subsystems cgroups.Hierarchy
 }
 
-func newCGroupsFetcher(hostRoot, cgroup string) *cgroupsFetcher {
+func newCGroupsFetcher(hostRoot, cgroup, mountsFilePath string) *cgroupsFetcher {
 	if cgroup != "" {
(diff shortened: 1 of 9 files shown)


src/raw/cgroup.go
```diff
@@ -16,22 +16,87 @@ import (
...
 
-func newCGroupsFetcher(hostRoot string) *cgroupsFetcher {
-	path := containerToHost(hostRoot, localCgroupPath)
+func newCGroupsFetcher(hostRoot, cgroup string) *cgroupsFetcher {
+	if cgroup != "" {
+		// Prepend cgroup to have higher priority than the predefined paths.
+		localCgroupPaths = append([]string{cgroup}, localCgroupPaths...)
+	}
+
+	cgroupPath, err := detectCgroupPath(localCgroupPaths)
```
varas: it's usually a good practice to extract env-var from runtime workflow and pass their values as usual args to constructors or runtime functions

we could improve this on next iteration

varas:
Changelog update pls!

PR #22, opened 2020-03-18
Add support for fetching metrics from ECS Fargate containers

douglascamata:
This PR adds capability of fetching ECS Fargate container metrics and metadata from the Metadata V3 endpoint. 

* The endpoint's URL is fetched from the container's environment. This env var is put there automagically by AWS.
* Metrics are fetched and parsed using types directly imported from Docker.
* Took the freedom to refactor the Dockerfile to provide an easy way to build the image with a compiled binary of this integration.

## TODO

- [x] Fetch container's labels
- [x] Fetch container's attributes (container image, name, image id, command, status...)

## Open questions

Fargate mode is behind a command-line argument named `fargate`. I'm considering trying to auto-detect Fargate. It's not that easy because the same environment variable can be present in ECS with EC2 launch type.


src/raw/aws/fargate_fetcher.go
```diff
@@ -0,0 +1,274 @@
...
+func (e *FargateFetcher) InspectContainer(containerID string) (docker.Container, error) {
+	defer func() {
+		if err := e.containerStore.Save(); err != nil {
+			e.logger.Errorf("persisting previous metrics: %v", err)
+		}
+	}()
+	var taskResponse TaskResponse
+	var containerResponse ContainerResponse
+
+	// Try to load container from the cache cpuStore.
```
douglascamata: Upcoming typo fix.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG infra_image=newrelic/infrastructure:latest
+ARG infra_image=newrelic/infrastructure-bundle
```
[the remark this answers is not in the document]
douglascamata: This is more like a development Dockerfile... so I need to overwrite the nri-docker binary inside the bundle image. :/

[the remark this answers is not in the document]
douglascamata: @carlosroman wanted even to remove this Dockerfile completely, because it seems like the bundle image will be the way to run any integration.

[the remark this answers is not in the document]
alejandrodnm: ok :)

src/nri/fargate_sampler.go
```diff
@@ -0,0 +1,71 @@
...
+
+const clusterNameLabel = "com.amazonaws.ecs.cluster"
+
+// FargateSampler is a sampler to get container data from an HTTP endpoint.
+type FargateSampler struct {
+}
+
+// SampleAll will sample all the information this sampler can fetch and populate it in the given integration.
+func (f *FargateSampler) SampleAll(i *integration.Integration) error {
+	httpClient := &http.Client{Timeout: 5 * time.Second}
```
ardias: will 5 seconds be enough? maybe make it a bit higher, it's not something that will block everything from running

src/raw/aws/fargate_fetcher.go
```diff
@@ -0,0 +1,274 @@
...
+	// Try to load container from the cache store.
+	_, err := e.containerStore.Get(containerID, &containerResponse)
+	if err == nil {
+		e.logger.Debugf("found container %s in cache cpuStore", containerID)
+		return containerResponseToDocker(containerResponse), nil
+	}
+
+	err = e.fetchTaskResponse(&taskResponse)
+	if err != nil {
+		return docker.Container{}, err
```
ardias: why not return a pointer here, like in some other places?

douglascamata: @ardias because this data structure is only used in the functions `func attributes(container types.Container)` and  `func labels(container types.Container) []entry`, both taking a valuei instead of pointer. These functions are also used in other parts of the code and I didn't want to change their signature.

src/nri/fargate_sampler.go
```diff
@@ -0,0 +1,72 @@
...
+type FargateSampler struct {
+}
+
+// SampleAll will sample all the information this sampler can fetch and populate it in the given integration.
+func (f *FargateSampler) SampleAll(i *integration.Integration) error {
+	httpClient := &http.Client{Timeout: fargateClientTimeout}
+	fetcher, err := aws.NewFargateFetcher(httpClient, i.Logger())
+	if err != nil {
+		return fmt.Errorf("could not create fargate fetcher: %v", err)
+	}
```
varas: can't we inject this service when constructing FargateSampler?

src/raw/aws/fargate_fetcher.go
varas: if the returned struct is not in a healthy state (therefore an empty value is expected by handlers) I rather returning a possible nil pointer so we can manage it

otherwise things like https://github.com/newrelic/nri-docker/pull/22/files#diff-d5c158e9c48d3c54469db640d9836a9eR63 could panic

don't take what's there as stone written as this integration was developed in a rush (you know beta mindset...)

varas: if the returned struct is not in a healthy state (therefore an empty value is expected by handlers) I rather returning a possible nil pointer so we can manage it

otherwise things like https://github.com/newrelic/nri-docker/pull/22/files#diff-d5c158e9c48d3c54469db640d9836a9eR63 could panic

don't take what's there as stone written as this integration was developed in a rush (you know beta mindset...)

src/biz/metrics_aws_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	assert.Equal(t, uint64(11), samples.Pids.Current)
+	assert.Equal(t, uint64(0), samples.Pids.Limit)
+
+	assert.Equal(t, float64(0), samples.BlkIO.TotalReadBytes)
+	assert.Equal(t, float64(0), samples.BlkIO.TotalReadCount)
+	assert.Equal(t, float64(7839744), samples.BlkIO.TotalWriteBytes)
+	assert.Equal(t, float64(957), samples.BlkIO.TotalWriteCount)
+
+	assert.Equal(t, Network{}, samples.Network)
```
douglascamata: Fargate has no network stats as of today, from my testing.

src/raw/aws/fargate_raw_metrics.go
```diff
@@ -0,0 +1,65 @@
...
+			continue
+		}
+		rawMetrics[containerID] = &raw.Metrics{
+			Time:        time.Time{},
+			ContainerID: containerID,
+			Memory: raw.Memory{
+				UsageLimit: stats.MemoryStats.Limit,
+				Cache:      stats.MemoryStats.Stats["cache"],
+				RSS:        stats.MemoryStats.Stats["rss"],
+				SwapUsage:  0,
```
douglascamata: Fargate doesn't return swap usage. It can be confirmed through the JSONs that are also part of this PR. They have the raw responses from the Metadata v3 API.

src/raw/cgroup.go
```diff
@@ -23,13 +23,16 @@ var localCgroupPaths = []string{
 	"/cgroup",
 }
 
-// cgroupsFetcher fetches the metrics that can be found in cgroups file system
-type cgroupsFetcher struct {
-	cgroupPath string
-	subsystems cgroups.Hierarchy
+// CgroupsFetcher fetches the metrics that can be found in cgroups file system
+// TODO: keep updating the code to use this fecher directly.
```
douglascamata: Removing this todo in next commit.

src/nri/sampler.go
```diff
@@ -28,42 +26,30 @@ const (
 type ContainerSampler struct {
 	metrics biz.Processer
 	store   persist.Storer
-	docker  dockerClient
+	docker  DockerClient
 }
 
-// abstraction of the docker client to facilitate testing
-type dockerClient interface {
+// DockerClient is abstraction of the Docker client.
```
douglascamata: Typo

douglascamata:
@carlosroman rebased, sir.

.travis.yml
```diff
@@ -1,9 +1,9 @@
 language: go
 
 go:
-  - "1.10"
+  - "1.14"
```
varas: I guess we centos5 is out of the scope :trollface:

Dockerfile
```diff
@@ -1,6 +1,6 @@
...
 
-FROM golang:1.10 as builder
+FROM golang:1.14 as builder
```
varas: I guess we centos5 is out of the scope :trollface:

src/biz/metrics.go
```diff
@@ -94,8 +94,9 @@ func (mc *MetricsFetcher) Process(containerID string) (Sample, error) {
 	if err != nil {
 		return metrics, err
 	}
+	// TODO: move logic to skip container without State to Docker specific code.
 	if json.State == nil {
-		return metrics, fmt.Errorf("invalid container %v JSON: missing State", json.ID)
+		log.Debug("invalid container %v JSON: missing State", json.ID)
```
varas: caller already knows about the container ID, so we could return a fixed error that's easier to handle by callers

varas: btw, why are you logging? shouldn't this prevent from fetching?

douglascamata: It doesn't prevent fetching at all in Fargate. So far, 100% of the Fargate containers I've seen don't have a `State`, they only have a `Status`. 

I want to keep the container ID in the log message to allow it to make sense without requiring any context from previous log messages. The implementation of the caller might change and then this log message would be incomplete.

src/biz/metrics_aws_test.go
varas: assert.Empty?

Dockerfile
douglascamata: Really? We cannot upgrade Go because of CentOS 5?

src/biz/metrics_aws_test.go
douglascamata: Fixed.

Dockerfile
varas: don't worry, it doesn't apply to this integration

src/nri/sampler.go
```diff
@@ -72,16 +58,18 @@ func NewSampler(hostRoot, cgroupPath string) (ContainerSampler, error) {
...
-	containers, err := cs.docker.ContainerList(context.Background(), types.ContainerListOptions{All: true})
-	if err != nil {
-		return err
-	}
 	defer func() {
 		if err := cs.store.Save(); err != nil {
 			log.Warn("persisting previous metrics: %s", err.Error())
 		}
 	}()
+
```
varas: 👏

src/nri/sampler.go
```diff
@@ -99,7 +87,13 @@ func (cs *ContainerSampler) SampleAll(i *integration.Integration) error {
 		populate(ms, attributes(container))
 		populate(ms, labels(container))
 
-		if container.State != "running" {
+		// TODO: this *needs* to be refactored into the call to ContainerList, because different
+		// systems might represent running container in a slightly different way. This can be tricky
+		// because for Docker containers we're relyng on the capabilities of the official Docker client.
+		// Possibly wrapping the Docker client with another type is a good solution.
+		// i.e. Docker uses `state = running` and ECS uses `status = RUNNING`.
```
varas: great comment, thanks!

src/raw/aws/fargate_metadata.go
```diff
@@ -0,0 +1,157 @@
...
+// metadataResponse gets the response from the given endpoint using the given HTTP client.
+func metadataResponse(client *http.Client, endpoint string) ([]byte, error) {
+	var resp []byte
+	var err error
+	for i := 0; i < maxRetries; i++ {
+		resp, err = sendMetadataRequest(client, endpoint)
+		if err == nil {
+			return resp, nil
+		}
+		fmt.Fprintf(os.Stderr, "Attempt [%d/%d]: unable to get metadata response from '%s': %v",
```
varas: log warning instead?

src/nri/sampler.go
```diff
@@ -72,16 +58,18 @@ func NewSampler(hostRoot, cgroupPath string) (ContainerSampler, error) {
...
-		return err
-	}
 	defer func() {
 		if err := cs.store.Save(); err != nil {
 			log.Warn("persisting previous metrics: %s", err.Error())
 		}
 	}()
+
+	// todo: configure to retrieve only the running containers
+	containers, err := cs.docker.ContainerList(context.Background(), types.ContainerListOptions{All: true})
```
varas: could we generate ctx in main? then pass it & handle properly?

douglascamata: I'm ok with generating the context in main.

You mean handling cancellation? If yes, I can this can be subject of another PR to avoid extra complications and increased scope here.

PR #24, opened 2020-03-23
Added LabelRename functionality to enable label normalization (for ECS for now)

jorikvdwerf:


src/nri/sampler_test.go
```diff
@@ -0,0 +1,112 @@
+package nri
```
carlosroman: ❤️ Thank you ❤️

PR #29, opened 2020-05-05
Fix cgroup path when cgroup driver is systemd

cristianciutea:
This PR will make the nri-docker integration compatible with systemd cgroup driver


src/docker.go
```diff
@@ -2,30 +2,30 @@ package main
...
+	CgroupDriver        string `default:"" help:"Optional. Specify the cgroup driver."`
+	DockerClientVersion string `default:"1.24" help:"Optional. Specify the version of the docker client. Used for compatibility."`
 }
 
 const (
-	integrationName     = "com.newrelic.docker"
-	integrationVersion  = "0.6.0"
-	dockerClientVersion = "1.24" // todo: make configurable
+	integrationName    = "com.newrelic.docker"
+	integrationVersion = "0.7.0"
```
carlosroman, line 28: This should be the next version released (`v1.3.0`).

jayv:
@cristianciutea @carlosroman any rough estimate when this could land in newrelic infra agent? 

We're running into this issue on a platform that is setup with `systemD`

PR #31, opened 2020-06-22
exec tini

xeago:
This makes tini 1, allowing it to effectively reap all zombies.


davidbrota:
Hi @xeago,

Thank you for having detected this issue. We changed the docker image some time ago and now it uses the `newrelic/infrastructure-bundle` as the base image. That image already calls tiny properly ([link](https://github.com/newrelic/infrastructure-agent/blob/84b36ab7c942b6ed6304733e882d98d5e47232e1/build/container/Dockerfile#L42). We'll remove this bash script as it's no longer in use.

Thanks

PR #32, opened 2020-07-16
Fix and simplify the condition to skip non-running containers

douglascamata:
The logic in this conditional was broken for Fargate containers, as the check on `container.Status != "running"` was removed.

I took some time to simplify the `if`, as it was very difficult to read. The logic to identify containers that are not running is:

* their **state** isn't `running` (case insensitive, covers classic cgroups containers)
* their **status** isn't `running` (case insensitive, covers ECS containers)
* their **status** doesn't start with `up` (case insensitive, covers cgroups container with the systemd driver)

There's no need to check if `state` or `status` are or not empty.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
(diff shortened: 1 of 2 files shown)


alejandrodnm:
Can you add the new behavior in the CHANGELOG

douglascamata:
@alejandrodnm I'll add the bugfix. In fact there should not be any new behavior.

PR #45, opened 2020-10-15
CI: migrate pipeline into GH Actions 

brushknight:
- [x] Add version print logic into codebase 
- [x] Add docker login for pull rate limitations 
- [x] Add snyk step as `needs` for prerelease once it will be ok
- [x] Migrate into Org secrets
- [x] Squash commits because Joan Porta did not accept CLA

note: no Windows/Darwin


.github/workflows/push_pr.yml
```diff
@@ -0,0 +1,95 @@
...
+    branches:
+      - main
+      - master
+  pull_request:
+
+env:
+  TAG: "v0.0.0" # not needed for go releaser, but kept for uniformity
+  REPO_FULL_NAME: ${{ github.event.repository.full_name }}
+  ORIGINAL_REPO_NAME: "newrelic/nri-docker"
+  SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
```
paologallinaharbur, line 14: we can scope snyk and we should use the org secret

vendor/vendor.json
```diff
@@ -26,6 +26,12 @@
 			"revision": "6c72808b55902eae4c5943626030429ff20f3b63",
 			"revisionTime": "2019-08-06T19:59:04Z"
 		},
+		{
```
paologallinaharbur, line 29: why a dependency was added?

paologallinaharbur: is it needed for the tests?

paologallinaharbur: makes sense

PR #52, opened 2020-11-17
Add feature to stop reporting exited containers after a certain TTL

douglascamata:
The Docker daemon might keep exited containers indefinitely if they're not properly removed by whoever started them: either by running such containers with `--rm`, by removing the container manually after its execution or by running `docker container prune`. 

To avoid spamming New Relic with this data this PR changes the behavior of this integration to avoid reporting containers that are exited for more than 24 hours ago. This value is configurable through the `ExitedContainersTTL` argument.

If some users want to keep the old behavior, they can set it to `0`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Changed
+* The integration by default does not report exited containers that are older than 24 hours.
+  This value can be configured using the `ExitedContainersTTL` argument using any value that 
+  can be parsed into a `time.Duration`, i.e. `1s`, `1m`, `1h`.
+  To replicate the old behavior of the integration, set this argument to `0` (zero).
+
 ## 1.3.3 (2020-11-12)
 ### Changed
 * Add metadata to samples from Fargate (#50)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 * nri-docker will no longer report containers that have been stopped for more than 24 hours.
-  This value can be configured using the `ExitedContainersTTL` argument using any value that 
-  can be parsed into a `time.Duration`, i.e. `1s`, `1m`, `1h`.
-  To replicate the old behavior of the integration, set this argument to `0` (zero).
+  This value can be configured using the `EXITED_CONTAINERS_TTL` environment variable using 
+  any value that can be parsed into a `time.Duration`, i.e. `1s`, `1m`, `1h`.
+  To replicate the old behavior of the integration, set this environment variable to `0` (zero).
 
 ## 1.3.3 (2020-11-12)
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## 1.3.4 (2020-11-19)
+## 1.4.0 (2020-11-19)
 
 ### Changed
 * nri-docker will no longer report containers that have been stopped for more than 24 hours.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## Unreleased
+## 1.3.4 (2020-11-19)
 
 ### Changed
 * nri-docker will no longer report containers that have been stopped for more than 24 hours.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Changed
-* The integration by default does not report exited containers that are older than 24 hours.
+* nri-docker will no longer report containers that have been stopped for more than 24 hours.
   This value can be configured using the `ExitedContainersTTL` argument using any value that 
   can be parsed into a `time.Duration`, i.e. `1s`, `1m`, `1h`.
   To replicate the old behavior of the integration, set this argument to `0` (zero).
--- a/src/biz/metrics.go
+++ b/src/biz/metrics.go
@@ -118,7 +118,7 @@ func (mc *MetricsFetcher) Process(containerID string) (Sample, error) {
 		log.Debug("invalid container %s JSON: missing State", containerID)
 	}
 
-	if mc.exitedContainerTTL != 0 && strings.ToLower(json.State.Status) == "exited" {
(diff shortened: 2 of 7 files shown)


src/biz/metrics_test.go
```diff
@@ -159,7 +150,7 @@ func stress(t *testing.T, args ...string) (containerID string, closeFunc func())
 	t.Helper()
 
 	arguments := []string{
-		"run", "-d", "--rm",
```
douglascamata, line 162: `--rm` is not needed  and this function returns a "cleanup" function that is used to remove the container after it's not needed or the test finishes.

src/biz/metrics.go
```diff
@@ -95,15 +109,31 @@ func (mc *MetricsFetcher) Process(containerID string) (Sample, error) {
...
 	if json.ContainerJSONBase == nil {
 		return metrics, errors.New("empty container inspect result")
 	}
 
+	// TODO: move logic to skip container without State to Docker specific code.
 	if json.State == nil {
 		log.Debug("invalid container %s JSON: missing State", containerID)
 	}
 
+	if mc.exitedContainerTTL != 0 && strings.ToLower(json.State.Status) == "exited" {
```
gsanchezgavier, line 121: if json.State == nil this part should be skipped to avoid a nil ptr error right?

src/nri/sampler.go
```diff
@@ -72,6 +70,15 @@ func (cs *ContainerSampler) SampleAll(ctx context.Context, i *integration.Integr
 	}
 
 	for _, container := range containers {
+		metrics, err := cs.metrics.Process(container.ID)
+		if err != nil {
+			log.Error("error fetching metrics for container %v: %v", container.ID, err)
```
gsanchezgavier, line 75: if the error is due to expired container this error would be logged anyway , looks like this sshould be after the next if block

src/biz/metrics.go
```diff
@@ -95,15 +109,31 @@ func (mc *MetricsFetcher) Process(containerID string) (Sample, error) {
...
+		exitTimestamp, err := time.Parse(time.RFC3339Nano, json.State.FinishedAt)
+		if err != nil {
+			return metrics, fmt.Errorf("invalid finished_at timestamp for exited container %s: %s (%v)",
+				containerID,
+				json.State.FinishedAt,
+				err,
+			)
+		}
+		if time.Now().After(exitTimestamp.Add(mc.exitedContainerTTL)) {
```
roobre, line 130: Really minor nitpick
```suggestion
		if time.Since(exitTimestamp) > mc.exitedContainerTTL {
```

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Changed
+* The integration by default does not report exited containers that are older than 24 hours.
```
roobre, line 11: I have a slight preference to state that this behavior actually changed, even if it is already in the "changelog section". Feel free to touch up the wording as you like though.
```suggestion
* Since 1.3.4, nri-docker will not report containers that have been stopped for more than 24 hours.
```

src/nri/sampler.go
```diff
@@ -72,6 +70,15 @@ func (cs *ContainerSampler) SampleAll(ctx context.Context, i *integration.Integr
...
+			if _, ok := err.(*biz.ErrExitedContainerExpired); ok {
+				log.Debug("skipping exited container older than TTL")
```
roobre, line 77: I think `Info` might be a better fit, as this message could be useful for users and not just developers.
```suggestion
				log.Info("skipping exited container older than TTL")
```

roobre, line 77: `ErrExitedContainerExpired` should contain a more descriptive error message (including the container ID), right?
Additionally I think `Info` might be a better level, as end users can also benefit from seeing this.
```suggestion
			if exitError, ok := err.(*biz.ErrExitedContainerExpired); ok {
				log.Info(exitError.Error())
```

CHANGELOG.md
douglascamata: I didn't do this because I'm not sure next release will be `1.3.4`. It could be `1.4.0` because there's a new feature. Or you could consider it a breaking change and do `2.0.0`, or other things could be merged and then it's impossible to predict what's the next release version.

roobre: Agreed, `1.3.4` should better be a placeholder. I'll edit the sugestion.

src/nri/sampler.go
douglascamata: Just FYI, this might spam logs quite heavily over time with repeated log lines.

The error will contain all the possible information, it's created with:

```go
ErrExitedContainerExpired{fmt.Sprintf("container %s exited after TTL (%v), skipping", containerID, mc.exitedContainerTTL)}

File: src/nri/sampler.go (hunk 3 above)
Comment (douglascamata) [reply src/nri/sampler.go:75]: This could happen for other reasons too, it was already part of the code before.

File: src/biz/metrics.go (hunk 4 above)
Comment (douglascamata) [reply src/biz/metrics.go:130]: Ahhh, this reads much better!

File: src/nri/sampler.go (hunk 6 above)
Comment (roobre) [reply src/nri/sampler.go:77]: Let me create a new suggestion, as I think it is easy to understand each other that way :)

File: src/nri/sampler.go [hunk 7]
```diff
@@ -72,6 +70,15 @@ func (cs *ContainerSampler) SampleAll(ctx context.Context, i *integration.Integr
...
+				continue
+			}
+		}
```
roobre, line 80:
```suggestion
		if err != nil {
			if _, ok := err.(*biz.ErrExitedContainerExpired); ok {
				log.Debug(err.Error())
				continue
			}
			log.Error("error fetching metrics for container %v: %v", container.ID, err)
		}
```

This way, `ErrExitedContainerExpired` is logged only once, but with full detail.

douglascamata:
@roobre @gsanchezgavier all comments taken care of, guys! Ready for review again. 👍

gsanchezgavier (approved):
thanks

douglascamata:
image

roobre:
@douglascamata we'd like to request a final change from you, which is explicitly mentioning in the changelog which environment variable and value must be set by the user to restore the old behavior.

We have also decided that this will be a minor version change (feel free to bump the version as well), as we do not consider tracking stopped containers the main use case of this integration, but rather an unwanted side-effect.

douglascamata:
@roobre do I just bump the minor in the changelog or is there anything else that needs to be done regarding the version bump?

roobre:
Changelog and [here](https://github.com/newrelic/nri-docker/blob/8c8c5bb8efd35e8542f9de80c9962a2bcdef6b33/src/docker.go#L29) should be enough :)

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## Unreleased
+## 1.3.4 (2020-11-19)
```
roobre, line 8: Please use 1.4.0, as while we do not consider the change breaking, it is still worth a minor version change. Thanks!

src/docker.go
```diff
@@ -28,7 +28,7 @@ type argumentList struct {
 
 const (
 	integrationName    = "com.newrelic.docker"
-	integrationVersion = "1.3.3"
+	integrationVersion = "1.3.4"
```
roobre, line 31: Same as above, `1.4.0` would be better :)

roobre (changes requested):

roobre (approved):
LGTM

PR #53, opened 2020-11-26
Fix bug in identifying exited containers and add more tests

douglascamata:


src/nri/sampler_test.go
```diff
@@ -48,6 +50,15 @@ func (m *mockStorer) Save() error {
 	return m.Called().Error(0)
 }
 
+type mockFecher struct {
```
roobre, line 53: s/Fecher/Fe**t**cher

src/nri/sampler_test.go
```diff
@@ -111,3 +122,80 @@ func TestECSLabelRename(t *testing.T) {
 		}
 	}
 }
+
+func TestExitedContainerTTL_Expired(t *testing.T) {
```
roobre, line 126:
```suggestion
func TestExitedContainerTTLExpired(t *testing.T) {
```

src/nri/sampler_test.go
```diff
@@ -111,3 +122,80 @@ func TestECSLabelRename(t *testing.T) {
...
+			Names:   []string{"Container 1"},
+			Image:   "my_image",
+			ImageID: "my_image_id",
+		},
+	}, nil)
+	mocker.On("ContainerInspect", mock.Anything, mock.Anything).Return(types.ContainerJSON{
+		ContainerJSONBase: &types.ContainerJSONBase{
+			State: &types.ContainerState{
+				Status:     "exited",
+				FinishedAt: time.Now().Add(-2 * time.Hour).Format(time.RFC3339Nano),
```
roobre, line 140: I'd probably add some more margin here just to be in the safe side, like:
```suggestion
				FinishedAt: time.Now().Add(-2 * time.Hour - time.Minute).Format(time.RFC3339Nano),
```
We don't want a leap second to ruin our test :)

src/nri/sampler_test.go
```diff
@@ -111,3 +122,80 @@ func TestECSLabelRename(t *testing.T) {
...
+			Names:   []string{"Container 1"},
+			Image:   "my_image",
+			ImageID: "my_image_id",
+		},
+	}, nil)
+	mocker.On("ContainerInspect", mock.Anything, mock.Anything).Return(types.ContainerJSON{
+		ContainerJSONBase: &types.ContainerJSONBase{
+			State: &types.ContainerState{
+				Status:     "exited",
+				FinishedAt: time.Now().Add(15 * time.Minute).Format(time.RFC3339Nano),
```
roobre, line 175: This would cause the `FinishedAt` timestamp to be in the future. While it should work just fine for this case, I don't think it's the case we want to test
```suggestion
				FinishedAt: time.Now().Add(-15 * time.Minute).Format(time.RFC3339Nano),
```

roobre (changes requested):

roobre:
Looks good to me, other than the small comments above. I have manually checked the coverage report and the error checking path is now covered 👍

src/nri/sampler_test.go
douglascamata: Hahahahaha, it won't! I'm using 30 mins as TTL in the test. 😂

roobre: > Hahahahaha, it won't! I'm using 30 mins as TTL in the test. 😂

That's cheating!! Hahha

Jokes aside it is fine to me in that case :)

roobre (approved):

roobre:
Let's wait a bit to see if @gsanchezgavier sees something that I might have missed :)

gsanchezgavier (approved):
thanks for adding the tests

roobre:
Feel free to merge @douglascamata, I'll do the release and run the jenkins pipeline afterwards :)

PR #58, opened 2021-03-22
fix: add correct config to the bundle

gsanchezgavier:
During the pipeline migration the config file added to the bundle was not the expected by the agent to be automatically activated. This Pr fixes that.


roobre (approved):
lgtm

Issue #34, opened 2020-08-03
Feature Request: Fargate Platform 1.4.0 Support


blytheaw:
Fargate Platform 1.4.0 adds a new v4 task metadata endpoint which provides some additional metrics in addition to what the v3 task metadata endpoint provides. Specifically, it adds network performance metrics such as bytes in/out and packet loss. This would be great to see added to the New Relic Docker integration to better support Fargate users.

In looking at the documentation, it seems the requests/response structure are the same for v4 endpoints, it just adds additional data in the response. However, it requires using the ECS_METADATA_URI_V4 environment variable to send requests to.

josemore:
thanks for submitting this @blytheaw 
Increased support for Fargate on EKS is something that it's planned so I'm adding this feedback for consideration. 
Though we can't commit on a timeline for resolution, I'll post an update here when we make any progress on this topic.

paologallinaharbur:
We are still in the process to release this

PR #60, opened 2021-03-30
feat(fargate): adding networking info

paologallinaharbur:
Adding support for[ Task metadata endpoint v4](https://docs.aws.amazon.com/AmazonECS/latest/userguide/task-metadata-endpoint-v4-fargate.html)

```
The task metadata endpoint version 4 functions like the version 3 endpoint,
but provides additional network metadata for your containers and tasks. 
Additional network metrics are available when querying the /stats endpoints as well.
```

### DIFF
--- a/src/biz/metrics_aws_test.go
+++ b/src/biz/metrics_aws_test.go
@@ -34,7 +34,6 @@ func TestFargateMetrics(t *testing.T) {
 	samples, err := metrics.Process(fargateContainerID)
 	require.NoError(t, err)
 
-	assert.Equal(t, float64(2), samples.CPU.LimitCores)
 	assert.Equal(t, uint64(200704), samples.Memory.CacheUsageBytes)
 	assert.Equal(t, uint64(268435456), samples.Memory.MemLimitBytes)
 	assert.Equal(t, uint64(11292672), samples.Memory.RSSUsageBytes)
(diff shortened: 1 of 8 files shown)


paologallinaharbur:
Fix https://github.com/newrelic/nri-docker/issues/34

paologallinaharbur:
It should fix as well https://github.com/newrelic/nri-ecs/issues/3, I am currently checking if the scope is exactly the same

src/docker.go
```diff
@@ -68,13 +69,18 @@ func main() {
 	var fetcher raw.Fetcher
 	var docker nri.DockerClient
 	if args.Fargate {
-		metadataV3BaseURL, err := aws.MetadataV3BaseURL()
+		var err error
+		var metadataBaseURL *url.URL
+		if metadataBaseURL, err = aws.MetadataV4BaseURL(); err != nil {
+			//If we do not find V4 we fall back to V3
```
gsanchezgavier, line 75: i'm a bit worried about masking aws.MetadataV4BaseURL() errors here. 
perhaps adding a debug log here with printing the error could solve that, wdyt?

src/raw/aws/fargate_metadata.go
```diff
@@ -157,3 +158,17 @@ func MetadataV3BaseURL() (*url.URL, error) {
...
+// MetadataV4BaseURL returns the v4 metadata endpoint configured via the ECS_CONTAINER_METADATA_URI environment
+// variable.
+func MetadataV4BaseURL() (*url.URL, error) {
+	baseURL, found := os.LookupEnv(containerMetadataEnvVarV4)
+	if !found {
+		return nil, fmt.Errorf("could not find env var with Metadata V4 API URL: %s", containerMetadataEnvVarV4)
+	}
+	parsedURL, err := url.Parse(baseURL)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse Metadata V4 API URL (%s): %s", baseURL, err)
```
gsanchezgavier, line 171: for instance if v4 fails because of this error , we are going to use v3 and not see any error

src/biz/metrics_aws_test.go
```diff
@@ -52,6 +51,47 @@ func TestFargateMetrics(t *testing.T) {
 	assert.Empty(t, samples.Network)
 }
 
+func TestFargateMetricsV4(t *testing.T) {
```
gsanchezgavier, line 54: you might want to have a look at [this test](https://github.com/newrelic/infrastructure-agent/blob/dca424e1edc3779e06aa68c51ef5c792811c2d84/test/databind/fargate/fargate_inside_test.go) , perhaps we can implement something similar based on the aws mock container

gsanchezgavier (approved):
lgtm , just an small suggestion on logs

src/docker.go
paologallinaharbur: you are right, I'll add a log

src/biz/metrics_aws_test.go
paologallinaharbur: I added an issue for that, the problem is that since nri-docker does not compile locally creating tests is "time consuming". I guess this is way there are few unit tests in the repo

src/raw/aws/fargate_metadata.go
```diff
@@ -157,3 +158,17 @@ func MetadataV3BaseURL() (*url.URL, error) {
 	}
 	return parsedURL, nil
 }
+
+// MetadataV4BaseURL returns the v4 metadata endpoint configured via the ECS_CONTAINER_METADATA_URI environment
+// variable.
+func MetadataV4BaseURL() (*url.URL, error) {
+	baseURL, found := os.LookupEnv(containerMetadataEnvVarV4)
+	if !found {
+		return nil, fmt.Errorf("could not find env var with Metadata V4 API URL: %s", containerMetadataEnvVarV4)
```
davidbrota, line 167: this error should be treated as a warning

davidbrota (commented):
LGTM, I agree with @gsanchezgavier comments about logging

PR #64, opened 2021-04-09
feat: add missing metrics from cgroup

### LINKED ISSUE #61: Add missing metrics from cgroups
Update github.com/containerd/cgroups library.

Add the following metrics:
- Docker CPU Shares
- Docker KMEM Usage
- Docker Mem Swap
- Docker Mem SW Limit
- Docker Mem Soft Limit
- Docker Mem SW In Use

gsanchezgavier:
This PR includes:
- Adds the following metrics:
`cpuShares`
`memoryKernelUsageBytes`
`memorySwapUsageBytes`
`memorySwapOnlyUsageBytes`
`memorySwapLimitBytes`
`memorySwapLimitUsagePercent`
`memorySoftLimitBytes`
- Changes how `memoryUsageBytes` and `memoryUsageLimitPercent` are calculated
## BREAKING
All swap memory has been removed from `memoryUsageBytes` and `memoryUsageLimitPercent`

Having the follow container:
```
docker run --memory-swap=400m --memory=300m --memory-reservation=250m stress stress-ng --vm 1 --vm-bytes 350m
```
We will get the following metrics (bytes transformed for simplicity of the example):
```
"memoryCacheBytes": 0.3mb,
"memoryResidentSizeBytes": 298.4 m,
"memorySizeLimitBytes": 300 m,

"memorySoftLimitBytes": 250 m,
"memorySwapLimitBytes": 400 m,
"memorySwapLimitUsagePercent": 88.27 %,
"memorySwapUsageBytes": 354.83 m,
"memorySwapOnlyUsageBytes": 54.83 m,

"memoryUsageBytes": 298.4 m,
"memoryUsageLimitPercent": 99.42 %,
```
`memoryUsageBytes` = `memoryResidentSizeBytes`
`memoryUsageLimitPercent` = `memoryUsageBytes` / `memorySizeLimitBytes`
`memorySwapLimitUsagePercent` = `memorySwapUsageBytes /  `memorySwapLimitBytes`

The reasoning behind this change is to follow the next convention and avoid confusion:
* Metrics with no swap reference in the name have no swap components
* Metrics with swap reference have memory+swap unless the contrary is specified like in memorySwapOnlyUsageBytes

Closes #61


src/biz/metrics.go
```diff
@@ -234,12 +242,30 @@ func (mc *MetricsFetcher) memory(mem raw.Memory) Memory {
 		usagePercent = 100 * float64(usage) / float64(memLimits)
 	}
 
+	swapLimit := mem.SwapLimit
+	if mem.SwapLimit > math.MaxInt64/2 {
+		swapLimit = 0
+	}
+	swapUsagePercent := float64(0)
+	if swapLimit > 0 {
+		swapUsagePercent = 100 * float64(mem.SwapUsage+mem.RSS) / float64(swapLimit)
```
roobre, line 251: Why are we adding `mem.RSS` to `mem.SwapUsage`? I'd expect those to be unrelated.

gsanchezgavier: Good point It's a bit confusing I will add a comment and perhaps I should change the metric name of swapLimit to add more clarity

src/raw/cgroup.go
```diff
@@ -270,6 +277,7 @@ func memory(metric *cgroups.Metrics) (Memory, error) {
 	mem.Cache = metric.Memory.Cache
 	mem.RSS = metric.Memory.RSS
 	mem.SwapUsage = metric.Memory.Swap.Usage
-
+	mem.SwapLimit = metric.Memory.HierarchicalSwapLimit
```
roobre, line 280: I think we are not being consistent with the use of `HierarchicalSwapLimit`. For getting the memory limit we use `Memory.Usage.Limit` rather than `HierarchicalMemoryLimit`, so we should just pick one of the two approaches and go with it.

Regarding which of the two is more correct, I'm not entirely sure. From my understanding, hierarchical limits are not enabled by default in containerd (not sure about Docker), and while one could use them to, for example, limit the amount of memory that a group of containers (or all containers in the system) can claim, I do not see this as common usage.

gsanchezgavier: good catch thanks! yes this is inconstant, I will use  `metric.Memory.Swap.Limit` then herarchical can be added if needed with explicit reference in the name

src/biz/metrics.go
```diff
@@ -217,29 +230,50 @@ func (mc *MetricsFetcher) memory(mem raw.Memory) Memory {
...
+	if mem.SwapUsage != 0 { // for systems that have swap disabled
+		swapOnlyUsage = mem.SwapUsage - mem.FuzzUsage
 	}
+	swapUsage := mem.RSS + swapOnlyUsage
 
-	usagePercent := float64(0)
-	if memLimits > 0 {
-		usagePercent = 100 * float64(usage) / float64(memLimits)
+	swapLimit := mem.SwapLimit
+	if mem.SwapLimit > math.MaxInt64/2 {
```
davidbrota, line 245: this is a nitpick, but I would compare against MaxInt32

davidbrota (approved):

src/biz/metrics.go
gsanchezgavier: that would be a different number , this is equivalent to `math.maxUint64/4`  `4294967296 Gb` . But i couldn't find the documentation that mention this on cgroups. But doing test i seen the memory limits exceed this number when is not set inside the cgroup file.

roobre (approved):
lgtm

PR #65, opened 2021-04-14
feat: add devicemapper storage metrics

### LINKED ISSUE #62: Add missing metrics from Docker API
Add docker client library github.com/docker/docker/client

Add following metrics:
- Docker Data Total
- Docker Data Percent

gsanchezgavier:
This PR add the Storage Driver metrics. This are parsed from the `/info` url ( you can see a similar output executing the cli docker info). 

Devicemapper is the only supported storage (since is the currently only exposing the data metrics).

metrics added:
`StorageDataUsedBytes`
`StorageDataAvailableBytes`
`StorageDataTotalBytes`
`StorageDataUsagePercent`
`StorageMetadataUsedBytes`
`StorageMetadataAvailableBytes`
`StorageMetadataTotalBytes`
`StorageMetadataUsagePercent`

Note: This information is structured in a key-value array and is not considered stable by docker according to the [API docs](https://docs.docker.com/engine/api/v1.41/#operation/SystemInfo).

Fixes #62


src/nri/sampler.go
```diff
@@ -69,6 +69,17 @@ func (cs *ContainerSampler) SampleAll(ctx context.Context, i *integration.Integr
 		return err
 	}
 
+	//TODO Add ff to get this info
```
davidbrota, line 72: I would add a config option to disable the collection of these metrics.

gsanchezgavier: Forgot to address that , thanks! I can add a config file to disable this metric collection.

gsanchezgavier:
rebased from base

src/config/config.go
```diff
@@ -0,0 +1,15 @@
+package config
```
davidbrota, line 1: 👏

davidbrota (approved):
LGTM

src/config/config.go
gsanchezgavier: i needed to import this struct in other packages.

src/biz/storage.go
```diff
@@ -0,0 +1,64 @@
...
+			stats.DataAvailable = value
+		case "Data Space Total":
+			stats.DataTotal = value
+		case "Metadata Space Used":
+			stats.MetadataUsed = value
+		case "Metadata Space Available":
+			stats.MetadataAvailable = value
+		case "Metadata Space Total":
+			stats.MetadataTotal = value
+		}
```
roobre, line 53: I'm not sure if other files in this package do it, but perhaps it could be a good idea to put a log.Verbose statement in the `default` branch so we log something when encountering unexpected metrics here.

gsanchezgavier: This is example of the key:values data here (from docker info)
```
 Storage Driver: devicemapper
  Pool Name: docker-259:1-1049197-pool
  Pool Blocksize: 65.54kB
  Base Device Size: 10.74GB
  Backing Filesystem: xfs
  Udev Sync Supported: true
  Data file: /dev/loop0
  Metadata file: /dev/loop1
  Data loop file: /var/lib/docker/devicemapper/devicemapper/data
  Metadata loop file: /var/lib/docker/devicemapper/devicemapper/metadata
  Data Space Used: 879MB
  Data Space Total: 107.4GB
  Data Space Available: 28GB
  Metadata Space Used: 18.64MB
  Metadata Space Total: 2.147GB
  Metadata Space Available: 2.129GB
  Thin Pool Minimum Free Space: 10.74GB
  Deferred Removal Enabled: true
  Deferred Deletion Enabled: true
  Deferred Deleted Device Count: 0
  Library Version: 1.02.146-RHEL7 (2018-01-22)
  ```
I thought it was going to be too verbose if we print from default since there are some metrics that will follow on this category always.
I would say that is expected to encounter unexpected metrics that is why i didn't log them

File: src/nri/sampler.go [hunk 4]
```diff
@@ -37,7 +39,7 @@ type DockerClient interface {
 }
 
 // NewSampler returns a ContainerSampler instance.
-func NewSampler(fetcher raw.Fetcher, docker DockerClient, exitedContainerTTL time.Duration) (*ContainerSampler, error) {
+func NewSampler(fetcher raw.Fetcher, docker DockerClient, exitedContainerTTL time.Duration, config config.ArgumentList) (*ContainerSampler, error) {
```
roobre, line 42: I don't like the idea of having such a high-level concept as the argument list passed this down to the code. Have you considered any alternatives, e.g. modeling this new functionality outside of the `Inspector`? Could perhaps be something like `Informer` with the `Info` method, and pass a nil one `if $FARGATE || cs.config.DisableStorageMetrics`

gsanchezgavier: I though to convert the argument struct to a config struct sanitized but there is nothing to sanitize , and i thought it was better to keep it simple. 
I could add optional parameters and add the bool parameter in the fetcher perhaps.
Let me have a look about the informer approach

gsanchezgavier: I reviewed the solution and all docker api calls are managed by DockerClient interface so it would be a bit confusing to call the info method of the api from other object i think.

roobre (approved):

PR #87, opened 2022-04-14
Cgroups v1 integration test mocked filesystem

alvarocabanas:
- Integration test with a mocked filesystem to check the output is the expected.


src/raw/cgroup.go
```diff
@@ -206,6 +206,7 @@ const nanoSecondsPerSecond = 1e9
 // statistics line and then sums up the first seven fields
 // provided. See `man 5 proc` for details on specific field
 // information.
+// TODO: we should inject the route of /proc/stat in order to be able to mock the file and test the method
```
sigilioso, line 209: As discussed, maybe we can take the chance to create a new issue to address the refactor to make this _testable_.

alvarocabanas: Yes, I will add this as an issue and I will add another issue with my first findings on needed refactors. For sure we will find many other on the process.

tests/integration/informer_mock_test.go
```diff
@@ -0,0 +1,29 @@
...
+type InspectorMock struct{}
+
+func (i InspectorMock) ContainerInspect(_ context.Context, _ string) (types.ContainerJSON, error) {
+	return types.ContainerJSON{
+		ContainerJSONBase: &types.ContainerJSONBase{
+			ID: containerID,
+			State: &types.ContainerState{
+				Pid: pid,
+			},
+			RestartCount: 2,
```
sigilioso, line 22: I would set `containerID`, `pid` and `RestartCount` values as `InspectorMock` attributes, I think the mock would be more flexible.

tests/integration/informer_mock_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"context"
+	"github.com/docker/docker/api/types"
+)
+
+const (
+	containerID = "my-container"
+	pid         = 666
+)
```
sigilioso, line 11: Nit: as those constants are defined for the whole `tests/integration` package I would be a bit more verbose with the naming.
```go
const (
	inspectorDefaultContainerID = "my-container"
	inspectorDefaultPid                = 666
)
```

tests/integration/informer_mock_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+type InspectorMock struct{}
```
sigilioso, line 13: Nit: I noticed the struct name and the file names don't match. Should it be inspector or informer?

sigilioso (commented):
Great job with the integration test 👍 
I've just let some small comments, mostly _nits_.

tests/integration/metrics_test.go
```diff
@@ -228,3 +234,160 @@ func TestExitedContainersWithoutTTL(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// CgroupsFetcherMock is the raw CgroupsFetcher with mocked cpu.systemUsage and time
+	// The hostRoot is our mocked filesystem
+	cgroupFetcher, err := NewCgroupsFetcherMock(hostRoot, "cgroupfs", "")
+	require.NoError(t, err)
+
+	storer := generateInMemoryStorerWithPreviousCPUState()
+
+	t.Run("Given a mockedFilesystem and previous CPU state Then processed metrics are as expected", func(t *testing.T) {
```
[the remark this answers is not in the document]
alvarocabanas: I just added it to to have a description of the test being done, it can be removed, we don't have a Standard for the tests. If this would be a Table Test there would be one Run for each execution. I used to do them with different in Run blocks to have a description and actually having all the setup on a Given Block, with the execution on a When Block and the assertions on a Then Block. But here I just wanted to have here the description of the test to not have a superLong Function name.

tests/integration/cgroups_fetcher_mock_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+// Fetch calls the wrapped fetcher and overrides the CPU.SystemUsage and the Time
+func (cgf *CgroupsFetcherMock) Fetch(c types.ContainerJSON) (raw.Metrics, error) {
+	metrics, err := cgf.cgroupsFetcher.Fetch(c)
+	if err != nil {
+		return raw.Metrics{}, err
+	}
+
+	metrics.CPU.SystemUsage = 19026130000000
+	metrics.Time = mockedTime
```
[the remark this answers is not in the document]
alvarocabanas: Yes that's true. I will do as you say.

tests/integration/informer_mock_test.go
alvarocabanas: It should be Inspector, I guess I had a short circuit 😕  I'll change it

tests/integration/metrics_test.go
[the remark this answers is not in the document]
sigilioso: 👍 thanks for the explanation! I think it can be a good way to give some context.
As there are plenty of ways to do go tests, I just wanted to know if we are always grouping tests like this or something like that.
As you said, another alternative is having a longer function name and maybe using a description in the `assert` call, but I also like this approach.

tests/integration/metrics_test.go
```diff
@@ -228,3 +234,167 @@ func TestExitedContainersWithoutTTL(t *testing.T) {
 		assert.Empty(t, sample)
 	})
 }
+
+const (
+	relativePathToTestdataFileystem = "testdata/cgroupsV1_host/"
+	InspectorContainerID            = "my-container"
+	InspectorPID                    = 666
```
gsanchezgavier, line 241: 👹

tests/integration/metrics_test.go
```diff
@@ -228,3 +234,167 @@ func TestExitedContainersWithoutTTL(t *testing.T) {
...
+			TotalWriteCount: 89,
+			TotalReadBytes:  2387968,
+			TotalWriteBytes: 50,
+		},
+		CPU: biz.CPU{
+			CPUPercent:       0.15003804623431827,
+			KernelPercent:    1.19999999,
+			UserPercent:      100,
+			UsedCores:        2.8546433726,
+			LimitCores:       float64(runtime.NumCPU()),
```
gsanchezgavier, line 273: nit: i guess that this metric is not collected from the files right? a comment might help to the future us to help understand it

tests/integration/metrics_test.go
```diff
@@ -228,3 +234,167 @@ func TestExitedContainersWithoutTTL(t *testing.T) {
...
+	require.NoError(t, err)
+
+	err = generateMockedProcMountsFile(cgroupsFolder, hostRoot)
+	require.NoError(t, err)
+
+	err = generateMockedProcPIDCGroupFile(hostRoot)
+	require.NoError(t, err)
+
+	err = genereateMockedProcNetDevFile(hostRoot)
+	require.NoError(t, err)
```
gsanchezgavier, line 318: nit: perhaps it would clean up the test code if this gets moved to another func WDYT?
It is a bit hard to understand what is part of the mock creation and what is the test itself.

tests/integration/metrics_test.go
```diff
@@ -228,3 +234,167 @@ func TestExitedContainersWithoutTTL(t *testing.T) {
...
+	// CgroupsFetcherMock is the raw CgroupsFetcher with mocked cpu.systemUsage and time
+	// The hostRoot is our mocked filesystem
+	cgroupFetcher, err := NewCgroupsFetcherMock(hostRoot, "cgroupfs", "", mockedTimeForAllMetricsTest, 19026130000000)
+	require.NoError(t, err)
+
+	storer := generateInMemoryStorerWithPreviousCPUState()
+	inspector := NewInspectorMock(InspectorContainerID, InspectorPID, 2)
+
+	t.Run("Given a mockedFilesystem and previous CPU state Then processed metrics are as expected", func(t *testing.T) {
+		metrics := biz.NewProcessor(storer, cgroupFetcher, inspector, 0)
```
gsanchezgavier, line 329: nit: I see that test is using exported methods only (i think) so it might be a good opportunity to change this file package to `biz_test`

tests/integration/metrics_test.go
```diff
@@ -228,3 +234,167 @@ func TestExitedContainersWithoutTTL(t *testing.T) {
...
+
+		sample, err := metrics.Process(InspectorContainerID)
+		require.NoError(t, err)
+		assert.Equal(t, expectedSample, sample)
+	})
+}
+
+// generateInMemoryStorerWithPreviousCPUState creates a storere with a previous CPU state
+// in order to make the processor calculate a Cpu Delta
+func generateInMemoryStorerWithPreviousCPUState() persist.Storer {
```
gsanchezgavier, line 339: nit: i think the generate prefix could be removed from the name

tests/integration/metrics_test.go
```diff
@@ -228,3 +234,167 @@ func TestExitedContainersWithoutTTL(t *testing.T) {
...
+	var previous struct {
+		Time int64
+		CPU  raw.CPU
+	}
+
+	storer := persist.NewInMemoryStore()
+	// We set the time as 10 seconds before the timestamp for the metrics
+	previous.Time = mockedTimeForAllMetricsTest.Add(-time.Second * 10).Unix()
+	previous.CPU = raw.CPU{
```
gsanchezgavier, line 348: just out of curiosity, the delta calculation doesn't work if previous values are 0 ?

tests/integration/metrics_test.go
```diff
@@ -228,3 +234,167 @@ func TestExitedContainersWithoutTTL(t *testing.T) {
...
+cgroup <HOST_ROOT>/pids cgroup rw,nosuid,nodev,noexec,relatime,pids 0 0
+cgroup <HOST_ROOT>/cpu,cpuacct cgroup rw,nosuid,nodev,noexec,relatime,cpu,cpuacct 0 0
+cgroup <HOST_ROOT>/cpuset cgroup rw,nosuid,nodev,noexec,relatime,cpuset 0 0
+cgroup <HOST_ROOT>/memory cgroup rw,nosuid,nodev,noexec,relatime,memory 0 0
+`
+	mountsContent = strings.ReplaceAll(mountsContent, "<HOST_ROOT>", cgroupsFolder)
+	return os.WriteFile(filepath.Join(hostRoot, "proc", "mounts"), []byte(mountsContent), 0755)
+}
+
+func generateMockedProcPIDCGroupFile(hostRoot string) error {
```
gsanchezgavier, line 374: couldn't this and the NetDevFile be with the others files in `/testdata` . I guess we should need to 'fix' the InspectorPID but could simplify the code a lot i think.

gsanchezgavier (commented):
thanks for adding extra tests I left some nits and questions

tests/integration/informer_mock_test.go
alvarocabanas: done

tests/integration/informer_mock_test.go
alvarocabanas: done

tests/integration/informer_mock_test.go
alvarocabanas: done

tests/integration/cgroups_fetcher_mock_test.go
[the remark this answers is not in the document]
alvarocabanas: done

tests/integration/metrics_test.go
```diff
@@ -263,7 +263,7 @@ func TestAllMetricsPresent(t *testing.T) {
 			KernelPercent:    1.19999999,
 			UserPercent:      100,
 			UsedCores:        2.8546433726,
-			LimitCores:       1,
+			LimitCores:       2,
```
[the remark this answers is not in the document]
alvarocabanas: It is coming from runtime but it was changed in another commit.

tests/integration/metrics_test.go
alvarocabanas: 🤘🏻

tests/integration/cgroups_fetcher_mock_test.go
```diff
@@ -0,0 +1,38 @@
+package biz
```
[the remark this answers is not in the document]
alvarocabanas: This biz package was already here, and there is no benefit from it, this biz package is living completely in the tests/integration folder and has no access to the biz package in the source, if you see, in the test the calls to that package are done biz.blabla... so this package can be called as we want because it will always be indepenedent.

tests/integration/metrics_test.go
alvarocabanas: The comment is on the the mock, but I will also add it here ok.

alvarocabanas: done

tests/integration/metrics_test.go
alvarocabanas: Same as above.

tests/integration/metrics_test.go
alvarocabanas: They are in the testdata, but here they are just copied to the specific folder where the code is looking at but done in the hostRoot that is the tempDir.

alvarocabanas: so yes, this files are already in the testdata.

tests/integration/metrics_test.go
alvarocabanas: done

tests/integration/cgroups_fetcher_mock_test.go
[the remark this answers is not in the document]
gsanchezgavier: > this biz package is living completely in the tests/integration folder and has no relation nor access to the biz package in the source

I totally miss this , cool

tests/integration/metrics_test.go
alvarocabanas: done

sigilioso (approved):
👍

sigilioso:
You can update from master to include [this](https://github.com/newrelic/nri-docker/pull/90) and remove the unsuccessful checks.

src/raw/cgroup.go
alvarocabanas: done

PR #89, opened 2022-04-14
Go modules

alvarocabanas:
Switch to go modules


go.mod.old
```diff
@@ -0,0 +1,33 @@
+module github.com/newrelic/nri-docker
+
+go 1.16
+
+require (
+	github.com/Azure/go-ansiterm v0.0.0-20170929234023-d6e3b3328b78 // indirect
```
paologallinaharbur, line 6: this should not be here I guess

alvarocabanas: Yes, I will delete it. I don't know why it was there.

alvarocabanas: done!

sigilioso (approved):
LGTM, I think we all are going to avail of this change soon 👍 . On the other hand, I think it would be good to have an extra review from somewhere with more experience in this integration.

sigilioso:
You can update from master to include [this](https://github.com/newrelic/nri-docker/pull/90) and remove the unsuccessful checks.

PR #93, opened 2022-04-22
refactor: [BREAKING CHANGE] stop supporting cgroupspath and cgroupsdriver configuration

sigilioso:
## Possible BREAKING CHANGE

Although the removed has been _marked as deprecated_ long ago and the removed configuration arguments don't appear either in documentation or in examples, those are two configuration parameters being removed. Therefore, any user relying on them could be affected.

### DIFF
--- a/src/config/config.go
+++ b/src/config/config.go
@@ -5,10 +5,8 @@ import "github.com/newrelic/infra-integrations-sdk/args"
 type ArgumentList struct {
 	args.DefaultArgumentList
 	HostRoot              string `default:"" help:"If the integration is running from a container, the mounted folder pointing to the host root folder"`
-	CgroupPath            string `default:"" help:"Optional. The path where cgroup is mounted."`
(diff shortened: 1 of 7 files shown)


alvarocabanas (approved):
LGTM, just wonder:  what happens if you still pass CgroupPath and/or CgroupDriver, I think the SDK just doesn't care about them when marshaling, but just in case, because, for example, with Viper you have the option of failing if extra params are there.

sigilioso:
> LGTM, just wonder: what happens if you still pass CgroupPath and/or CgroupDriver, I think the SDK just doesn't care about them when marshaling, but just in case, because, for example, with Viper you have the option of failing if extra params are there.

Good point, I was thinking that extra fields would be ignored, but after taking a deeper look, it is actually easy to make it fail

```
go run ./src -cgroup_path=xxx -cgroup_driver=xxx
flag provided but not defined: -cgroup_path
```

Therefore, I'm keeping the config fields for backwards compatibility (even though the fields will be completely ignored).

Thank you for pointing it out!

sigilioso:
Parameters kept for backwards compatibility, @alvarocabanas could you review it again, please?

PR #94, opened 2022-04-25
refactor: raw cgroups fetcher factory

sigilioso:


src/raw/cgroupv1.go
```diff
@@ -0,0 +1,264 @@
+// Package raw fetches raw system-level metrics as they are presented by the operating system
+package raw
```
alvarocabanas, line 2: NIT: WDYT if we place the cgroupsV1 stuff in a folder and package called cgroupsv1 and the fetcher would be cgroupsv1.Fetcher, then the same for cgroupsV2. I'm not sure if it makes sense or there are too many interdependent stuff and the actual structure is better, but I want to open the discussion.

alvarocabanas (approved):
LGTM, but I left a NIT to discuss something about the structure.

src/raw/cgroupv1.go
sigilioso: That thought also crossed my mind but I would rather keep just one package. Although implementations are different, both are still related to the same. Besides, being pragmatic, I think there will be some code which will be common for v1 and v2... Then, having two separate packages would complicate things a little bit, wouldn't it?

PR #97, opened 2022-04-27
cgroups detect v2

sigilioso:
Introduces capabilities to check cgroups version and driver and to obtain `cgroups2` mount-point and group.

In order to detect cgroup driver and version we needed to call `docker info` through docker API before cgroups fetcher creation and that info was already needed for specific use-cases in the integration. So, a new docker-client wrapper is being included to avoid repeating the `docker info` request.

As a result, some refactoring was needed and the PR became bigger than expected. Although changes are not very complicated, the PR may be easier to review commit by commit.


src/raw/cgroupv2_detect.go
```diff
@@ -0,0 +1,125 @@
...
+		return fmt.Sprintf(cgroupV2SystemdTemplate, containerID), nil
+	case CgroupGroupfs:
+		return fmt.Sprintf(cgroupV2CgroupfsTemplate, containerID), nil
+	}
+	return "", fmt.Errorf("invalid cgroup2 driver %q", driver)
+}
+
+// cgroupV2MountPoint returns the cgroup mount point which is built joining info from mountsFile (Eg: /proc/mounts)
+// and pid's cgroup file (Eg: /proc/<pid>/cgroup)
+func cgroupV2MountPoint(hostRoot string, pid int, fileOpen fileOpenFn) (string, error) {
```
sigilioso, line 55: I am not proud of this implementation because it has a lot of code from `cgroupv1_detect` duplicated. I think we should check if this can be avoided, but doing it now (by changing the cgroupv1 detect implementation too) can be more risky than doing it once the cgroup2 support feature is more advance. So, I created the #99 issue so we do not forget it.

tests/integration/cgroups_fetcher_mock_test.go
```diff
@@ -18,7 +18,7 @@ type CgroupsFetcherMock struct {
 
 // NewCgroupsFetcherMock creates a new cgroups data fetcher.
 func NewCgroupsFetcherMock(hostRoot string, time time.Time, systemUsage uint64) (*CgroupsFetcherMock, error) {
-	cgroupsFetcher, err := raw.NewCgroupsFetcher(hostRoot)
+	cgroupsFetcher, err := raw.NewCgroupsV1Fetcher(hostRoot)
```
marcsanmi, line 21: Do you think it woud make sense this mock to support `CgroupsV2Fetcher` as well? 🤔 Anyway, this can be addressed when implementing the tests for `cgroupsV2Fetcher`.

marcsanmi (approved):
LGTM!

tests/integration/cgroups_fetcher_mock_test.go
sigilioso: As you said, we can find out the best approach once we implement `cgroupsV2Fetcher` tests. At the moment, I chose to use the `v1` builder instead of the generic one because only the `v1` is being tested.

alvarocabanas (approved):
LGTM, I just see that maybe when we have finished all the raw refactor and creation of new components, we can give a second thought if creating more packages inside the raw package to give more domain expressivity to the whole solution. This raw/biz separation it's not so clear from the first beginning.

sigilioso:
> LGTM, I just see that maybe when we have finished all the raw refactor and creation of new components, we can give a second thought if creating more packages inside the raw package to give more domain expressivity to the whole solution. This raw/biz separation it's not so clear from the first beginning.

Thanks, I couldn't agree more, current structure is quite confusing (biz, raw and raw/aws), we could create a _nice to have_ issue to review and possibly modify it.

PR #98, opened 2022-04-27
Add fetcher for containerd cgroupsv2

marcsanmi:
This PR implements a new fetcher using the containerd library cgroups v2.

Apparently there are 4 metrics that aren't available in cgroups v2 that are available in v1 (annotated with `TODO`  in the codebase):
- `per_cpu`
-  `blkio.throttle.io_service_bytes`
-  `blkio.throttle.io_serviced`
-   `cpu_shares`

We can do further investigation, or simply leave as it is... 

**Note:** Another different issue is regarding the new set of metrics that comes with v2 that are not included in our metrics.


src/raw/cgroupv2.go
```diff
@@ -0,0 +1,156 @@
...
+	// TODO: missing metric
+	//cgroupFullPathBlkio, err := cgroupInfo.getFullPath(cgroups.Blkio)
+	//
+	//if err == nil {
+	//	if stats.Blkio, err = blkio(cgroupFullPathBlkio); err != nil {
+	//		log.Error("couldn't read blkio stats: %v", err)
+	//	}
+	//} else {
+	//	log.Error("couldn't read blkio stats: %v", err)
+	//}
```
sigilioso, line 79: Nit: the _TODO_ and commented code are remainders of the blkio missing metric. As you has already made it explicit in the PR description, maybe we can get rid of this comment. WDYT?

src/raw/cgroupv2.go
```diff
@@ -0,0 +1,156 @@
...
+
+// Fetch get the metrics that can be found in cgroups v2 file system
+// Unlike v1, cgroup v2 has only single hierarchy.
+//TODO: populate also network from libcgroups
+func (cg *CgroupsV2Fetcher) Fetch(c types.ContainerJSON) (Metrics, error) {
+	stats := Metrics{}
+
+	pid := c.State.Pid
+	containerID := c.ID
+	driver := c.Driver
```
sigilioso, line 38: The `c.Driver` values is not the cgroups driver, that value can be got from `cgroupsInfo`. I'm afraid we would need to change the fetcher a little bit to retrieve and keep the cgroups driver value.

sigilioso (commented):
Changes are really good overall 👍 

I left a comment which I think should be addressed and an small nit

src/raw/cgroupv2.go
marcsanmi: Sure 👍

src/raw/cgroupv2.go
marcsanmi: You are right, updated in 1abb753b88863734b1b9ee2a35f19e5c46c1b207.

sigilioso (approved):
Thanks for addressing the comments!

LGTM 👍

src/raw/cgroupv2.go
```diff
@@ -0,0 +1,146 @@
...
+	if metric.Memory == nil {
+		return mem, errors.New("no Memory metrics information")
+	}
+	if metric.Memory.Usage != 0 {
+		mem.UsageLimit = metric.Memory.UsageLimit
+		mem.FuzzUsage = metric.Memory.Usage
+	}
+
+	mem.Cache = metric.Memory.File
+	mem.RSS = metric.Memory.Anon
```
alvarocabanas, line 128: In the table created to show the mapping of each value in NR, cgroupsV1 and cgroupsV2 
It's saying that the memvory.rss is memory.swap_limit I gues it's a typo on that table, isn't it? Because here I see the RSS metric comes from Memory.Anon

alvarocabanas (approved):
LGTM, I've just added a question

src/raw/cgroupv2.go
marcsanmi: Yes, it was actually a typo. Already updated, thanks!

PR #101, opened 2022-05-04
include io metrics for cgroup2

sigilioso:


src/raw/metrics.go
```diff
@@ -6,6 +6,11 @@ import (
 	"github.com/docker/docker/api/types"
 )
 
+const (
```
marcsanmi, line 9: Should we define these constants in the `cgroupv2.go` where they are used?

marcsanmi (approved):
LGMT, just left a nit.

src/raw/metrics.go
marcsanmi: Liked this reasoning 👍

alvarocabanas (approved):
LGTM

Issue #104, opened 2022-05-05
[nri-docker] review coreLimitCores metric for cgroups1 and fargate


sigilioso:
After including `coreLimitCores` and `cpuPercent` metrics support for cgroups2 (see #84), we may use the same approach for cgroups1 (getting values from [cpuset](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v1/cpusets.html)) and fargate (honor cpu count got from API).

We would need to check possible implications / impact and perform the change.

sigilioso:
As @marcsanmi and I discussed, It is important to take into account `raw.CPU.OnlineCPUs` and `biz.CPU.LimitCores` meaning can be different. Therefore it could make sense to keep the current `biz.CPU.LimitCores` for cgroup1 and fargate and change it to have the same meaning in cgroup2.

PR #105, opened 2022-05-05
cgroups v2 cpus count

sigilioso:


src/raw/aws/fargate_raw_metrics.go
```diff
@@ -36,7 +36,6 @@ func fargateRawMetrics(fargateStats FargateStats) map[string]*raw.Metrics {
 				ThrottledPeriods:  stats.CPUStats.ThrottlingData.ThrottledPeriods,
 				ThrottledTimeNS:   stats.CPUStats.ThrottlingData.ThrottledTime,
 				SystemUsage:       stats.CPUStats.SystemUsage,
-				OnlineCPUs:        uint(stats.CPUStats.OnlineCPUs),
```
sigilioso, line 39: We were not using this value before, so after the needed changes in [biz/metrics.go](https://github.com/newrelic/nri-docker/pull/105/files#diff-f904c524386409b976146175f99fcb7ed6b5d7e450a5de0e3cc361cbc6e0ac88R166), the way I found to keep the previous behaviour is to stop defining `OnlineCPUs` value here.

I created the issue #104 to review if we would rather prefer to honor the `OnlineCPUs` value from the API in fargate.

marcsanmi (approved):
LGTM!

src/raw/cgroup.go
```diff
@@ -8,3 +16,49 @@ func NewCgroupsFetcher(hostRoot string, cgroupInfo *CgroupInfo, systemCPUReader
 
 	return NewCgroupsV1Fetcher(hostRoot, systemCPUReader)
 }
+
+func countCpusetCPUsFromPath(path string) (uint, error) {
```
marcsanmi, line 20: Nit: Ara these both funcs just cgroupV2-specific? If so, I was thinking on moving them inside cgroupv2

PR #106, opened 2022-05-05
Refactor net dev stats getter

alvarocabanas:
Refactor the Ned dev stats retrieval to use a struct that will be used by both fetchers.


marcsanmi (commented):
LGTM until now. You need to update the tests as you modified the fetcher constructor.

sigilioso:
Looks good! 👍  I'll review it again when the fetcher builder thing is fixed and new unit-tests added. BTW, ping me if you need help resolving conflicts.

PR #113, opened 2022-05-09
fix: cpus count and units

sigilioso:


marcsanmi (dismissed):

marcsanmi (commented):
Left and right of the "interval" should have been included, LGTM when the integration test is updated.

marcsanmi (approved):
👍

PR #114, opened 2022-05-09
Refactor cgroups detectors

alvarocabanas:
- Refactor cgroups detector to have same logic for both cgroups and be able to mock detector.

### DIFF
--- a/src/raw/cgroup_detect.go
+++ b/src/raw/cgroup_detect.go
@@ -1,17 +1,119 @@
 package raw
 
 import (
+	"bufio"
+	"fmt"
 	"io"
 	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/newrelic/infra-integrations-sdk/log"
 )
 
 const (
-	mountsFilePath    = "/proc/mounts"
-	cgroupFilePathTpl = "/proc/%d/cgroup"
+	mountsFilePath           = "/proc/mounts"
+	cgroupFilePathTpl        = "/proc/%d/cgroup"
+	cgroup1MountName         = "cgroup"
+	cgroup2MountName         = "cgroup2"
+	cgroup2UnifiedFilesystem = "/"
 )
+
+type CgroupDetector interface {
+	PopulatePaths(hostRoot string, pid int) error
+}
 
 type fileOpenFn func(string) (io.ReadCloser, error)
 
 func defaultFileOpenFn(filePath string) (io.ReadCloser, error) {
 	return os.Open(filePath)
+}
+
+func getMountsFile(hostRoot string, mountPoints map[string]string, cgroupMountPointName string, fileOpen fileOpenFn) error {
+	path := filepath.Join(hostRoot, mountsFilePath)
+	mountsFile, err := fileOpen(path)
+	if err != nil {
+		return fmt.Errorf("failed to open file: %s, while detecting cgroup mountpoints error: %v",
+			mountsFilePath, err)
+	}
+
+	defer func() {
+		if closeErr := mountsFile.Close(); closeErr != nil {
+			log.Error("Error occurred while closing the file: %v", closeErr)
+		}
+	}()
+
+	sc := bufio.NewScanner(mountsFile)
+	for sc.Scan() {
+		line := sc.Text()
+		fields := strings.Fields(line)
+
+		switch cgroupMountPointName {
+		case cgroup1MountName:
+			// Filter mount points if the type is not 'cgroup' or not mounted under </host>/sys
+			if len(fields) < 3 || !strings.HasPrefix(fields[2], cgroup1MountName) || !strings.HasPrefix(fields[1], hostRoot) {
+				continue
+			}
+
+			for _, subsystem := range strings.Split(filepath.Base(fields[1]), ",") {
+				if _, found := mountPoints[subsystem]; !found {
+					mountPoints[subsystem] = filepath.Dir(fields[1])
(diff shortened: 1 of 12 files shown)


src/raw/cgroupv1.go
```diff
@@ -38,20 +49,18 @@ func (cg *CgroupsV1Fetcher) Fetch(c types.ContainerJSON) (Metrics, error) {
...
-		err        error
-	)
-	cgroupInfo, err = getCgroupV1Paths(cg.hostRoot, pid)
-
+	err := cg.cgroupDetector.PopulatePaths(cg.hostRoot, pid)
 	if err != nil {
 		return stats, err
 	}
 
+	cgroupInfo := cg.cgroupDetector.(*CgoupsV1Detector).paths
```
sigilioso, line 57: As I see it, `CgroupDetector` interface could be a good idea because it would allow us to mock and uninttest fetchers. However, v1 and v2 detectors results are quite different and, as a result, castings are needed. For this reason, I think we do not get benefit from `CgroupDetector` abstraction. Furthermore, If we tried to implement a mock by implementing current `CgroupDetector` we would not be able to use it in fetchers because we are already casting to the specific type. 

Therefore, I would keep references to the specific structs instead, something like this for cgroupv1:

```go
// pathsDetector would have current *CgroupsV1Detector type, we may improve naming
paths, err := cg.pathsDetector.Paths()
```

If we decide to mock this in order to test fetchers, then I would implement the abstraction by defining different interfaces for `*CgroupsV1Detector` and `*CgroupsV2Detector`.

What do you think?

src/raw/network_stats_getter.go
```diff
@@ -3,14 +3,35 @@ package raw
...
+			"couldn't fetch network stats for container %s from cgroups: %v",
+			containerID,
+			err,
+		)
+	}
+	return net, err
+}
+
+// network fetches the network metrics from the /proc file system
+// TODO: use cgroups library + split this to open and parse.
```
sigilioso, line 33: Maybe we could take the chance to address the _split this to open and parse_ part of this TODO?

src/raw/cgroupv1.go
marcsanmi, line 57: Maybe I'd feel more confortable if we check for the value in the type assertion, wdyt?

src/raw/cgroup_detect.go
```diff
@@ -1,17 +1,119 @@
...
+	"github.com/newrelic/infra-integrations-sdk/log"
 )
 
 const (
-	mountsFilePath    = "/proc/mounts"
-	cgroupFilePathTpl = "/proc/%d/cgroup"
+	mountsFilePath           = "/proc/mounts"
+	cgroupFilePathTpl        = "/proc/%d/cgroup"
+	cgroup1MountName         = "cgroup"
+	cgroup2MountName         = "cgroup2"
```
marcsanmi, line 18: Nit: for consistency with the rest of the codebase I will rename them to something like `cgroupv1.*` and `cgroupv2.*`

src/raw/cgroup_detect.go
```diff
@@ -1,17 +1,119 @@
...
+			if fields[0] == "0" && fields[1] == "" {
+				cgroupPaths[cgroup2UnifiedFilesystem] = fields[2]
+				return nil
+			}
+		default:
+		}
+	}
+
+	return sc.Err()
+}
```
sigilioso, line 119: Great! You managed to stop repeating the whole files reading logic! 😄
I thought in some kind of iterator but I like this approach more because there are just two possible implementations.

src/raw/cgroup_detect.go
```diff
@@ -1,17 +1,119 @@
...
+				if _, found := mountPoints[subsystem]; !found {
+					mountPoints[subsystem] = filepath.Dir(fields[1])
+				}
+			}
+		case cgroup2MountName:
+			if len(fields) >= 3 && strings.HasPrefix(fields[2], cgroup2MountName) && strings.HasPrefix(fields[1], hostRoot) {
+				mountPoints[cgroup2UnifiedFilesystem] = fields[1]
+				return nil
+			}
+		default:
```
sigilioso, line 68: Nit: could we get rid of the empty default block? (the same in `getCgroupFilePaths` function)

src/raw/cgroupv1.go
alvarocabanas: You are right.
I have implemented interfaces per CgroupsVersion, both for the detector and the paths, I think now it gets much better.

src/raw/network_stats_getter.go
alvarocabanas: I will create an issue to address this and add a test in a separate PR, WDYT?

src/raw/cgroupv1.go
alvarocabanas: This is now changed and it's not needed anymore.

alvarocabanas: The integrations are now separated per version

src/raw/cgroup_detect.go
alvarocabanas: Done

src/raw/cgroup_detect.go
alvarocabanas: done

sigilioso (approved):
Awesome work! Thank you for addressing the comments 👍

marcsanmi (approved):
LGTM 👍

PR #117, opened 2022-05-11
Network stats getter test

alvarocabanas:
Refactor network_stats_getter and system_cpu_reader to have similar structure with open and parse and add tests.


test/integration/metrics_test.go
```diff
@@ -246,7 +246,7 @@ func TestExitedContainersWithoutTTL(t *testing.T) {
 		"/",
 		raw.NewCgroupV1PathParser(),
 		raw.NewPosixSystemCPUReader(),
-		raw.NetDevNetworkStatsGetter{},
+		raw.NewNetDevNetworkStatsGetter(),
```
sigilioso, line 249: This is out of the scope of this PR but I think we could use some kind of helper to build the testing fetcher (we are repeating the `raw.NewCgroupsV1Fetcher(...` call a lot.

sigilioso (approved):
LGTM 👍

test/integration/metrics_test.go
alvarocabanas: Yes it's true

marcsanmi (approved):
LGTM!

PR #120, opened 2022-05-11
Update CPU LimitCores metric

marcsanmi:
Closes https://github.com/newrelic/nri-docker/issues/104.

### Problem

This PR fixes the CPU metric `LimitCores` by honouring first the CPU quota if any, then checking for `OnlineCPUs` and  setting it to `runtime.NumCPU()` otherwise.

The problem with the new implementation was that by checking first `OnlineCPUs` that under the hood it's checking the `cpuset.cpus.effective` file, we didn't take into account the cpu quota, being more restrictive than the number of CPUs granted to the cgroup;  so, starting a docker container with `--cpus=0.5` and granted `2` cpus to the cgroup, the `LimitCores` metric would be `2` instead of `0.5`.

### Considerations

We need to make a decision about the old behavior and see if we want to implement the `LimitCores` metric following the same approach as in cgroupsv2 by checking for the [cpuset.effective_cpus](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v1/cpusets.html) instead of setting it right away to `runtime.NumCPU()`.

It's also worth mentioning that we have never taken into account the fact that regardless of whether there is CPU quota, it's never used as multiplier for the percentage metrics. The value that is used is the total available cpus, which seems a bit strange... 
e.g. Given `cpu=0.5` and available cpus `2` (from `cpus.effective` or `len(perCpuUsage)` depending on the cgroup version) the values are:
- `limitCores=0.5`
- multiplier used for percentage metrics `2*100`


src/biz/metrics_test.go
```diff
@@ -151,3 +154,89 @@ func TestCpuPercent(t *testing.T) {
...
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mc := &MetricsFetcher{store: store}
+
+			if got := mc.cpu(tt.args.cpu, tt.args.json); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("MetricsFetcher.cpu() = %v, want %v", got, tt.want)
```
sigilioso, line 238: Nit: as we are only comparing `LimitCores ` here, I would avoid using `reflect.DeepEqual`. What do you think about something like this?

```go
			if got := mc.cpu(tt.args.cpu, tt.args.json); got.LimitCores != want.LimitCores {
			    // ...
```

or using testify:

```go
assert.Equal(t, want.LimitCores, got.LimitCores)
```

sigilioso (approved):
👍 LGTM, I've just left a nit.

Maybe we should also check the behaviour with @mangulonr

src/biz/metrics_test.go
marcsanmi: I forgot to add it in the description... I did it this way as it would be easily extendable in case we want to check all the cpu metrics (which would be goo imo). We would need just to update the test name and add all the rest of metrics. If we want to do it separately you approach is by far easier. WDYT?

alvarocabanas (approved):
LGTM I just left a comment about the storer used in the test

src/biz/metrics_test.go
sigilioso: Well, we would also need to update the storer (to keep other previous values) and both the args and expectation to include the rest of the metrics, so I would rather keep it simple until we extended it. Anyway, it is not important.

marcsanmi:
We end up deciding to remove `OnlineCPUs` for `cpu.LimitCores` as it wasn't reflecting properly the metric.

src/biz/metrics.go
```diff
@@ -164,15 +164,13 @@ func (mc *MetricsFetcher) cpu(metrics raw.Metrics, json *types.ContainerJSON) CP
 
 	cpu := CPU{}
 
-	if metrics.CPU.OnlineCPUs != 0 {
-		cpu.LimitCores = float64(metrics.CPU.OnlineCPUs)
-	} else { // fallback if OnlineCPUs info is not available
+	// Set LimitCores to first honor CPU quota if any; otherwise try to set it from OnlineCPUs.
```
sigilioso, line 167:
```suggestion
```

The behaviour was changed but the comment was not, maybe we can get rid of it now, WDYT?

sigilioso (approved):
Thanks for addressing the last changes!

src/biz/metrics.go
```diff
@@ -81,6 +81,7 @@ type MetricsFetcher struct {
 	fetcher            raw.Fetcher
 	inspector          raw.DockerInspector
 	exitedContainerTTL time.Duration
+	getRuntimeNumCPU   func() int
 }
```
alvarocabanas, line 85: nice!

alvarocabanas (approved):
LGTM!

Issue #116, opened 2022-05-10
Update CPU and Memory integration tests


marcsanmi:
Update CPU and Memory integration tests to use the new fetcher constructor and make them work in both cgroups v1 and cgroups v2.

PR #121, opened 2022-05-12
Update cpu and memory tests to use generic fetcher constructor

marcsanmi:
Closes https://github.com/newrelic/nri-docker/issues/116.


test/integration/metrics_test.go
```diff
@@ -340,6 +311,30 @@ func TestAllMetricsPresent(t *testing.T) {
...
+// fetcher is a helper function that returns a Fetcher.
+//
+// If any error happens, the function will make to fail the given test.
+func fetcher(t *testing.T, docker *client.Client) raw.Fetcher {
+	t.Helper()
+
+	cgroupInfo, err := raw.GetCgroupInfo(context.Background(), docker)
+	if err != nil {
+		t.Fatalf("Creating CgroupInfo: %v", err)
+	}
```
sigilioso, line 323: Nit: since testify is already used in the project, you may want to use it to check errors (it is a bit easier to read).

```go
require.NoError(t, err)
```

sigilioso (approved):
👍  Thanks for addressing it, awesome job! Now the actual integration tests (the ones without the mocked filesystem) are ready to work in both cgroups1 and cgroups2.

I just left a small nit.

test/integration/metrics_test.go
marcsanmi: You're right, updated!

sigilioso:
👍

PR #128, opened 2022-05-23
Adding E2E tests cgroups v1

carlossscastro:
This PR adds E2E tests for docker with cgroups v1

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Compile nri-docker
         run: |
           GOOS=linux GOARCH=amd64 make compile
-      - name: Run E2E
+      - name: Run E2E for Cgroups v1
         uses: newrelic/newrelic-integration-e2e-action@v1
         with:
           spec_path: test/e2e/e2e_spec.yml
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,30 @@
+name: E2E
+
+on:
+  pull_request:
+
+env:
+  GO_VERSION: '1.18'
+
+jobs:
+  e2eTests:
+    # Do not run e2e tests if commit message or PR has skip-e2e.
+    if: ${{ ! contains(github.event.pull_request.labels.*.name, 'ci/skip-e2e') }}
+    runs-on: ubuntu-20.04
+    steps:
+      - name: checkout-repository
+        uses: actions/checkout@v2
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: Compile nri-docker
+        run: |
+          GOOS=linux GOARCH=amd64 make compile
+      - name: Run E2E
+        uses: newrelic/newrelic-integration-e2e-action@v1
+        with:
+          spec_path: test/e2e/e2e_spec.yml
+          account_id: ${{ secrets.COREINT_E2E_ACCOUNT_ID }}
+          api_key: ${{ secrets.COREINT_E2E_API_KEY }}
+          license_key: ${{ secrets.COREINT_E2E_LICENSE_KEY }}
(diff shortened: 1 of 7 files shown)


test/e2e/e2e_spec.yml
```diff
@@ -0,0 +1,39 @@
...
+        - query: "SELECT average(docker.container.cpuUsedCores) FROM Metric"
+      entities:
+        - type: "CONTAINER"
+          data_type: "Metric"
+          metric_name: "docker.container.cpuUsedCores"
+      metrics:
+        - source: "docker.yml"
+          except_metrics:
+            - docker.container.memorySoftLimitBytes
+            - docker.container.pids
```
marcsanmi, line 28: what's the problem with pids?

test/e2e/e2e_spec.yml
```diff
@@ -0,0 +1,39 @@
...
+            - docker.container.processCountLimit
+            - docker.container.storageDataAvailableBytes
+            - docker.container.storageDataTotalBytes
+            - docker.container.storageDataUsagePercent
+            - docker.container.storageDataUsedBytes
+            - docker.container.storageMetadataAvailableBytes
+            - docker.container.storageMetadataTotalBytes
+            - docker.container.storageMetadataUsagePercent
+            - docker.container.storageMetadataUsedBytes
+          # additionals: ""
```
marcsanmi, line 39: Could we remove this?

marcsanmi, line 39: Could we remove this?

marcsanmi (commented):
LGTM but I left a question and a nit.

test/e2e/e2e_spec.yml
carlossscastro: @marcsanmi 
From the samples I've seen `pids` are not frequently reported. As such I decided to exclude them from the list to avoid test failure. They don't seem to be a standard/basic metric

marcsanmi: It was out of curiosity, but seems we have never reported them... So better skip them 👍

marcsanmi (approved):
good job! 👍

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,30 @@
+name: E2E
+
+on:
+  pull_request:
+
+env:
+  GO_VERSION: '1.18'
+
+jobs:
+  e2eTests:
```
gsanchezgavier, line 10: nit: comment the cgroups version of the OS (i think is v1 but I would explicitly mentioned it for the future us) , might be as job name

test/e2e/agent_dir/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: '3.5'
+
+services:
+  agent:
+    build:
```
gsanchezgavier, line 5: perhaps the Docker file could be removed, and use directly the image here mounting the needed files as volumes. I think it would reduce complexity and execution time

test/e2e/agent_dir/newrelic-infra-agent/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM newrelic/infrastructure-bundle:latest
```
gsanchezgavier, line 1: why the bundle? i think `infrastructure` should be enough since the rest of the integrations are not needed

gsanchezgavier (approved):
looks good! i left some nits, thanks

.github/workflows/e2e.yml
carlossscastro: Nice. Changed!

test/e2e/agent_dir/newrelic-infra-agent/Dockerfile
carlossscastro: You're right, infra image is enough

test/e2e/agent_dir/docker-compose.yml
carlossscastro: Good catch. Changed on https://github.com/newrelic/nri-docker/pull/128/commits/d6a2322a75d686fd9dcfa71bed1c4c61fb4dc669

gsanchezgavier (approved):

PR #129, opened 2022-05-27
Add ci job to run e2e tests for cgroups v2

marcsanmi:
These are the four excluded metrics that aren't available:
- `docker.container.memorySizeLimitBytes`
- `docker.container.memoryUsageLimitPercent`
- `docker.container.memorySwapLimitBytes`
- `docker.container.memorySwapLimitUsagePercent`

All of them are directly consumed from the [containerd library](https://github.com/containerd/cgroups/blob/188f73d13086fdfc4a26bdda70f0362203552e3a/v2/manager.go#L452), related to memory limits. IMHO it makes sense as the resource limitations has changed from cgroups v1 and v2, including the resource limitation feature on a container basis in v2.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -31,7 +31,7 @@ jobs:
   e2eTests-cgroups-v2:
     # Do not run e2e tests if commit message or PR has skip-e2e.
     if: ${{ ! contains(github.event.pull_request.labels.*.name, 'ci/skip-e2e') }}
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     steps:
       - name: checkout-repository
         uses: actions/checkout@v2
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -7,7 +7,7 @@ env:
   GO_VERSION: '1.18'
 
 jobs:
-  e2eTests:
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e.yml
```diff
@@ -28,3 +28,24 @@ jobs:
           account_id: ${{ secrets.COREINT_E2E_ACCOUNT_ID }}
           api_key: ${{ secrets.COREINT_E2E_API_KEY }}
           license_key: ${{ secrets.COREINT_E2E_LICENSE_KEY }}
+  e2eTests-cgroups-v2:
+    # Do not run e2e tests if commit message or PR has skip-e2e.
+    if: ${{ ! contains(github.event.pull_request.labels.*.name, 'ci/skip-e2e') }}
+    runs-on: ubuntu-20.04
```
alvarocabanas, line 34: shouldn't it be ubuntu-22.04
?

sigilioso: I think so, with ubuntu-20.04 it is running with cgroups v1, isn't it?

marcsanmi: I do hate copy&paste 😆

sigilioso (approved):
👍 Thanks for addressing it!

PR #140, opened 2022-08-16
Update CHANGELOG.md

carlossscastro:
Bump dependencies


kang-makes (approved):
LGTM.

I would not even wait for the e2e to pass as the PR from the deps passed correctly.

PR #145, opened 2023-02-09
Fixes wrong calculation of OnlineCpus

alvarocabanas:
Fixes wrong calculation of OnlineCpus when the PercpuUsage array has more entries than CPUs


sigilioso (approved):
LBTM

PR #152, opened 2023-04-13
Remove attributes with empty values (NR-107455)

gsanchezgavier:
Depending on the source some of the attributes are not being populated, for instance in Fargate the `state` is not present on the metadata API. 

Currently the integration is sending this attributes with an empty value, this PR removes these attributes when the values is an empty string.

Note:
I studied the possibility to add this into the[ integration SDK](https://github.com/newrelic/infra-integrations-sdk/blob/v3.7.3/data/metric/metrics.go#L91) but since that func is returning an error it would be a breaking change, and also there might be a use case where empty values would make sense to be sent.

### DIFF
--- a/src/nri/sampler.go
+++ b/src/nri/sampler.go
@@ -165,11 +165,9 @@ func attributes(container types.Container) []entry {
 	// Removes attributes with emtpy values to avoid be reported.
 	sanitizedEntries := []entry{}
 	for _, entry := range entries {
-		if isAttributeValueEmpty(entry) {
-			continue
+                if !isAttributeValueEmpty(entry) {
+			sanitizedEntries = append(sanitizedEntries, entry)
 		}
-
-		sanitizedEntries = append(sanitizedEntries, entry)
 	}
 
 	return sanitizedEntries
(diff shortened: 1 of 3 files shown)


src/nri/sampler.go
```diff
@@ -153,14 +153,35 @@ func attributes(container types.Container) []entry {
...
+
+	return sanitizedEntries
+}
+
+func isAttributeValueEmpty(e entry) bool {
+	if e.Type == metric.ATTRIBUTE {
+		if strVal, ok := e.Value.(string); ok && strVal == "" {
+			return true
+		}
+	}
```
marcsanmi, line 183:
```suggestion
	if e.Type == metric.ATTRIBUTE {
             strVal, ok := e.Value.(string)
             return ok && strVal == ""
        }
```

marcsanmi (commented):
Thanks for addressing it!
LGTM, I added a couple of nits improving readability.

src/nri/sampler.go
```diff
@@ -153,14 +153,35 @@ func attributes(container types.Container) []entry {
...
+			continue
+		}
+
+		sanitizedEntries = append(sanitizedEntries, entry)
+	}
+
+	return sanitizedEntries
+}
+
+func isAttributeValueEmpty(e entry) bool {
```
paologallinaharbur, line 176: can we check if this would affect possibly empty values (if supported)

gsanchezgavier: I added some assert to the tests to check that labels are still being reported when they have empty values which is supported by docker.

PR #153, opened 2023-05-16
Add renovate automerge config

gsanchezgavier:


gsanchezgavier:
I would let renovate bump the docker dep with vuln after this gets merged

.github/renovate.json5
```diff
@@ -0,0 +1,25 @@
...
+    {
+      "matchUpdateTypes": ["minor", "patch"],
+      // Exclude any dependencies which are pre-1.0.0 because those can make breaking changes at any time
+      // according to the SemVer spec.
+      "matchCurrentVersion": "!/^(0|v0)/",
+      // Renovate will do the following when detecting a new dependency bump:
+      // - If it's one of the exclusions it will create a PR (major, 0. or v0. and sarama)
+      // - In case it matches one of the rest cases, it will create a branch that will be auto merged if tests pass.
+      // - If in the previous step tests fail, a PR will be created instead of merging the branch.
+      "automerge": true,
```
paologallinaharbur, line 20: do we need to add the coreint bot as an exception in the list of user that can push without a PR?

paologallinaharbur (approved):

.github/renovate.json5
gsanchezgavier: yes! that's already there thanks to the repo settings automation

gsanchezgavier: actually is not there 🤔  i will double check since it should

PR #174, opened 2023-07-13
fix: Use PRATE kind for perSecond metrics

gsanchezgavier:
In certain cases like when a restart happens the previous value of these metrics could be higher than the current which will produce a negative metric since the type is  `RATE`, this situation could bring confusion and averaged charts.

This PR modifies the type to `PRATE` which in case the rate computation is negative, [the metric is not sent](https://staging.onenr.io/0bEjOD8VpQ6).


paologallinaharbur (approved):
Thanks for tacking care of it :)

PR #188, opened 2023-11-15
Allow to set the cache TTL of storer (NR-179361)

kang-makes:
A customer states that is not able to see some CPU metrics if the interval is set to more than 1 minute.

These CPU Metrics that are failing rely on the storer to be able to calculate the delta/rate and the storer has hardcoded the value to 1 minute.

This PR allows the storer TTL to be variable for customers who might need it.

### DIFF
--- a/src/config/config.go
+++ b/src/config/config.go
@@ -13,4 +13,6 @@ type ArgumentList struct {
 	// CgroupPath and CgroupDriver arguments are not used but are kept here for backwards compatibility reasons.
 	CgroupPath   string `default:"" help:"Deprecated. cgroup_path argument is not used anymore."`
 	CgroupDriver string `default:"" help:"Deprecated. cgroup_driver argument is not used anymore."`
+
+	CacheTTL string `default:"1m" help:"Set the maximum cache that the integration is going to use to calculate rates and deltas. Possible values are time-strings: 1s, 1m, 1h"`
 }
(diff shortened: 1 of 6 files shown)


src/config/config.go
```diff
@@ -13,4 +13,6 @@ type ArgumentList struct {
 	// CgroupPath and CgroupDriver arguments are not used but are kept here for backwards compatibility reasons.
 	CgroupPath   string `default:"" help:"Deprecated. cgroup_path argument is not used anymore."`
 	CgroupDriver string `default:"" help:"Deprecated. cgroup_driver argument is not used anymore."`
+
+	CacheTTL string `default:"1m" help:"Set the maximum cache that the integration is going to use to calculate rates and deltas. Possible values are time-strings: 1s, 1m, 1h"`
```
sigilioso, line 17:
```suggestion
	CacheTTL string `default:"1m" help:"Set the maximum cache TTL that the integration is going to use to calculate rates and deltas. Possible values are time-strings: 1s, 1m, 1h"`
```

🤔 I guess this TTL should be greater than the interval. We will need to include some comments in the docs if that's correct.

sigilioso (approved):
Thanks for taking care of this!

sigilioso:
One of the integration tests is failing. Could you take a look in case it is related to the changes?

sigilioso (commented):
One of the integration tests is failing. Could you take a look in case it is related to the changes?

src/config/config.go
kang-makes: Sure.

kang-makes:
> One of the integration tests is failing. Could you take a look in case it is related to the changes?

Nope. The mock was broken. I fixed it so I ping you for review again @sigilioso :)

PR #197, opened 2024-02-01
chore: use reusable workflows

gsanchezgavier:
- Changes renovate config to skip bumping the runners used for cgroup v1 testing
- Adds reusable workflows

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -11,8 +11,14 @@ jobs:
   e2eTests-cgroups-v1:
     # Do not run e2e tests if commit message or PR has skip-e2e.
     if: ${{ ! contains(github.event.pull_request.labels.*.name, 'ci/skip-e2e') }}
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-20.04
     steps:
+      - name: Check cgroups version
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e.yaml
```diff
@@ -11,8 +11,14 @@ jobs:
   e2eTests-cgroups-v1:
     # Do not run e2e tests if commit message or PR has skip-e2e.
     if: ${{ ! contains(github.event.pull_request.labels.*.name, 'ci/skip-e2e') }}
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-20.04
```
marcsanmi, line 14: Why this? We were testing cgroups v1 with `ubuntu-22.04`, isn't it?

gsanchezgavier: no we weren't ,  I suspect that we merged a renovate Pr bumping this runner image , this is why I added the renovate exception and the extra check on the cgroups version docker has configured in the runner. Sooner or later this testing approach will be gone when that runner image gets removed.

gsanchezgavier: shame on me https://github.com/newrelic/nri-docker/pull/181/files

marcsanmi: I thought we were, good point 👍

marcsanmi (approved):
LGTM!

PR #201, opened 2024-02-07
NR-221180: docker api fetcher scaffolding

sigilioso:
## Docker API fetcher scaffolding

This PR introduces a new fetcher to get the metrics through the docker API. The fetcher is not used yet.

### Main changes
* The main function has been split to make it more readable and easier to extend.
* A new `DockerStatsClient` interface has been defined.
* A new package `raw.dockerapi` has been introduced, container the new fetcher scaffolding.
* A new temporal integration tests has been included: it shows how to build the new fetcher and how to fetch stats metrics from docker API. It can be used to quickly get docker container inspect and stats data in order to continue implementing the new fetcher.

### Notes for reviewers
* The new fetcher is not ready to be used (it includes some _unimplemented_ methods which would panic). However, since it is not used, the PR can be merged safely.
* Most metrics should come from `docker stats` but, if additional information was needed, the fetcher would be easy to extend.
* The `DockerAPIVersion` can be already configured using the integration arguments. The minimum required version and/or the default version might need to be adjusted in follow-up iterations (depending on the metrics needs).
* The new integration test was included for development purposes only, should be replaced by a proper integration test when the metrics are actually fetched.

### DIFF
--- a/src/docker.go
+++ b/src/docker.go
@@ -52,46 +52,61 @@ func main() {
 
 	var fetcher raw.Fetcher
 	var docker raw.DockerClient
+
 	if args.Fargate {
-		var err error
-		var metadataBaseURL *url.URL
-		if metadataBaseURL, err = aws.MetadataV4BaseURL(); err != nil {
-			log.Debug("The Metadata endpoint V4 is not available, falling back to V3: %s", err.Error())
-			//If we do not find V4 we fall back to V3
-			metadataBaseURL, err = aws.MetadataV3BaseURL()
-		}
-		exitOnErr(err)
-
-		fetcher, err = aws.NewFargateFetcher(metadataBaseURL)
-		exitOnErr(err)
-
-		docker, err = aws.NewFargateInspector(metadataBaseURL)
-		exitOnErr(err)
+		fetcher, docker = setupFargateFetcher()
 	} else {
-		var tmpDocker *client.Client
-		tmpDocker, err = client.NewClientWithOpts(client.FromEnv, client.WithVersion(args.DockerClientVersion))
-		exitOnErr(err)
-		defer tmpDocker.Close()
-		docker = raw.NewCachedInfoDockerClient(tmpDocker)
-
-		detectedHostRoot, err := raw.DetectHostRoot(args.HostRoot, raw.CanAccessDir)
-		exitOnErr(err)
-
-		cgroupInfo, err := raw.GetCgroupInfo(context.Background(), docker)
-		exitOnErr(err)
-
-		fetcher, err = raw.NewCgroupsFetcher(
-			detectedHostRoot,
-			cgroupInfo,
-			raw.NewPosixSystemCPUReader(),
-			raw.NewNetDevNetworkStatsGetter(),
-		)
-		exitOnErr(err)
+		fetcher, docker = setupCgroupsFetcher(args)
 	}
+
 	sampler, err := nri.NewSampler(fetcher, docker, args)
 	exitOnErr(err)
 	exitOnErr(sampler.SampleAll(context.Background(), i))
 	exitOnErr(i.Publish())
+}
+
+func setupFargateFetcher() (raw.Fetcher, raw.DockerClient) {
+	var err error
+	var metadataBaseURL *url.URL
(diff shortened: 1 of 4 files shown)


src/raw/dockerapi/fetcher.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func (f *Fetcher) Fetch(container types.ContainerJSON) (raw.Metrics, error) {
+	containerStats, err := f.ContainerStats(context.Background(), container.ID)
+	if err != nil {
+		return raw.Metrics{}, fmt.Errorf("could not fetch stats for container %s: %w", container.ID, err)
+	}
+	now := time.Now() // nolint: staticcheck
+	metrics := raw.Metrics{
+		Time:        now,
```
marcsanmi, line 28: Nit: we could just define it inline.
```suggestion
		Time:        time.Now(),
```

test/integration/dockerapi_fetcher_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"context"
+	"encoding/json"
+	"testing"
+
+	"github.com/docker/docker/client"
+	"github.com/newrelic/nri-docker/src/raw/dockerapi"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDockerAPIHelpers(t *testing.T) {
```
marcsanmi, line 13: We should extend it once the fetcher is fetching something.

marcsanmi (approved):
LGTM 👍

test/integration/dockerapi_fetcher_test.go
sigilioso: TODO added here: [a086ccd](https://github.com/newrelic/nri-docker/pull/201/commits/a086ccdf8fa3ce72d63bfc3841851b3af1bb60f4). Thanks! 🙂

src/raw/dockerapi/fetcher.go
sigilioso: I've updated it, although the `staticcheck` linter still complains if I remove the _nolint_:

```
src/raw/dockerapi/fetcher.go:27:16: SA4017: Now doesn't have side effects and its return value is ignored (staticcheck)
		Time:        time.Now(),
		             ^
```

🤔 I don't fully understand why it is complaining. I guess It could have something to do with the `Fetch` method not being called yet.

gsanchezgavier (approved):
thanks!

src/docker.go
```diff
@@ -52,48 +52,63 @@ func main() {
...
+	docker, err := aws.NewFargateInspector(metadataBaseURL)
+	exitOnErr(err)
+
+	return fetcher, docker
+}
+
+func setupCgroupsFetcher(args config.ArgumentList) (raw.Fetcher, raw.DockerClient) {
+	var err error
+	var dockerClient *client.Client
+	dockerClient, err = client.NewClientWithOpts(client.FromEnv, client.WithVersion(args.DockerClientVersion))
```
paologallinaharbur, line 90: why not 
```
dockerClient, err := client.NewClientWithOpts(client.FromEnv, client.WithVersion(args.DockerClientVersion))
```

sigilioso: I agree but I only wanted to move the code to a different (without changing the current implementation). Let's keep the change of of the scope of this if you don't mind.

src/docker.go
```diff
@@ -52,48 +52,63 @@ func main() {
...
 
+func setupFargateFetcher() (raw.Fetcher, raw.DockerClient) {
+	var err error
+	var metadataBaseURL *url.URL
+	if metadataBaseURL, err = aws.MetadataV4BaseURL(); err != nil {
+		log.Debug("The Metadata endpoint V4 is not available, falling back to V3: %s", err.Error())
+		// If we do not find V4 we fall back to V3
+		metadataBaseURL, err = aws.MetadataV3BaseURL()
+	}
+	exitOnErr(err)
```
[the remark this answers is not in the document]
sigilioso: I agree, but let's keep it out of the scope of this PR (we can improve the whole `docker.go` file in a different PR)

src/docker.go
paologallinaharbur: ah ok I did not notice it was already like that!

PR #203, opened 2024-02-12
fix: held release

sigilioso:
Held releases temporally, since #201 introduced a bug


gsanchezgavier (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 Unreleased section should follow [Release Toolkit](https://github.com/newrelic/release-toolkit#render-markdown-and-update-markdown).
 
+## Held
```
gsanchezgavier, line 10: not sure how long will be held, but in case is too long, perhaps we could revert the commit to avoid blocking sec updates

sigilioso: It is already identified, we'll fix it today

PR #206, opened 2024-02-12
feat: introduce dockerapi fetcher scaffolding

sigilioso:
This PR introduces the dockerapi fetcher scaffolding without the issues in #201


sigilioso:
> I would actually keep the old base since I was working on the top of your PR. Just to avoid rebases

Fair enough, closing this in favor of #207

PR #208, opened 2024-02-13
feat(main): refactor main

paologallinaharbur:
Changes:
 - cleaned up main:
   - extracted printVersion
   - split the Fargate and Docker data collection in two, having them "merged" made everything more complex
 - Removed the cache for the info. There was no reason to create a cache and a wrapped client, so I decided to store an Info result instead.
 - remove redundant parameters from NewCgroupsV1Fetcher and NewCgroupsV2Fetcher

**The code should be more readable, however, let me know if you prefer to have this PR as it is or split in 2/3**


src/config/config.go
```diff
@@ -5,11 +5,13 @@ import "github.com/newrelic/infra-integrations-sdk/args"
 type ArgumentList struct {
 	args.DefaultArgumentList
 	HostRoot              string `default:"" help:"If the integration is running from a container, the mounted folder pointing to the host root folder"`
+	Ecs                   bool   `default:"false" help:"If true and if cgoupsv2 are detected, metrics are collected through Docker API automatically changing the UseDockerAPI default to true"`
```
sigilioso, line 8:
```suggestion
	Ecs                   bool   `default:"false" help:"If true and metrics are collected through the Docker API when cgroupsv2 are detected, it has no effect otherwise"`
```

src/config/config.go
```diff
@@ -5,11 +5,13 @@ import "github.com/newrelic/infra-integrations-sdk/args"
...
 	Fargate               bool   `default:"false" help:"Enables Fargate container metrics fetching. If enabled no metrics are collected from cgroups or Docker. Defaults to false"`
 	ExitedContainersTTL   string `default:"24h" help:"Enables to integration to stop reporting Exited containers that are older than the set TTL. Possible values are time-strings: 1s, 1m, 1h"`
 	DockerClientVersion   string `default:"1.24" help:"Optional. Specify the version of the docker client. Used for compatibility."`
 	DisableStorageMetrics bool   `default:"false" help:"Disables storage driver metrics collection."`
 	ShowVersion           bool   `default:"false" help:"Print build information and exit"`
+	UseDockerAPI          bool   `default:"false" help:"Leverage Docker API to gather metrics instead of filesystem. This option is available for cgroupV2 only."`
```
sigilioso, line 14:
```suggestion
	UseDockerAPI          bool   `default:"false" help:"Leverage Docker API to gather metrics instead of filesystem. This option is available for cgroupsV2 only."`
```

sigilioso (approved):
LGTM

sigilioso:
As discussed _ofline_, let's check if it makes more sense to keep one flag only: `UseDockerAPI` (documenting that the flag will have no effect unless cgroupsv2 is used)

paologallinaharbur:
> As discussed ofline, let's check if it makes more sense to keep one flag only: UseDockerAPI (documenting that the flag will have no effect unless cgroupsv2 is used)
I'll add them in a follow up PR to avoid adding new functionality with this PR

marcsanmi (approved):
LGTM! simpler, cleaner 👍

src/raw/aws/fargate_metadata.go
```diff
@@ -172,3 +172,14 @@ func MetadataV4BaseURL() (*url.URL, error) {
 	}
 	return parsedURL, nil
 }
+
+// GetMetadataBaseURL returns metadataV4BaseURL if available, otherwise try with metadataV3BaseURL
+func GetMetadataBaseURL() (*url.URL, error) {
+	metadataBaseURL, err := metadataV4BaseURL()
```
gsanchezgavier, line 178: nit
```suggestion
	if metadataBaseURL, err := metadataV4BaseURL(); err==nil {
```

src/raw/aws/fargate_metadata.go
```diff
@@ -172,3 +172,14 @@ func MetadataV4BaseURL() (*url.URL, error) {
...
+	if err == nil {
+		return metadataBaseURL, nil
+	}
+	log.Debug("The Metadata endpoint V4 is not available, falling back to V3: %s", err.Error())
+	// If we do not find V4 we fall back to V3
+	return metadataV3BaseURL()
```
gsanchezgavier, line 184: both functions does exactly the same , perhaps this could be just one func with the env var as parameter , or even do the logic here IMO it doesn't have the complexity worth the layering

src/raw/client.go
```diff
@@ -28,24 +20,3 @@ type DockerClient interface {
 type DockerStatsClient interface {
 	ContainerStats(ctx context.Context, containerID string, stream bool) (types.ContainerStats, error)
 }
-
```
gsanchezgavier, line 31: 👍

src/raw/cgroup.go
```diff
@@ -6,20 +6,26 @@ import (
...
+	detectedHostRoot, err := DetectHostRoot(hostRoot, CanAccessDir)
+	if err != nil {
+		return nil, err
 	}
 
-	return NewCgroupsV1Fetcher(hostRoot, NewCgroupV1PathParser(), systemCPUReader, networkStatsGetter)
+	if cgroupInfo.CgroupVersion == CgroupV2 {
+		return NewCgroupsV2Fetcher(detectedHostRoot, cgroupInfo.CgroupDriver, NewPosixSystemCPUReader())
+	}
+	// if cgroupInfo.Version == CgroupV1
```
gsanchezgavier, line 27: looks like there is a reason to pick v1 by default , do you know if there is any corner-case when the API doesn't provide this info and is v1?

I guess this line is used in the following PR but just in case I saw the commented code.

gsanchezgavier (approved):
thanks!!

src/raw/aws/fargate_metadata.go
paologallinaharbur: done!

src/raw/cgroup.go
paologallinaharbur: I think that it is there for legacy reasons. We could error out since it should be impossible. But since we are not 100% sure we have that behavior

PR #209, opened 2024-02-13
Feat: Add memory metrics for docker api fetcher

marcsanmi:
This PR improves the Docker stats fetcher by implementing the `memoryMetrics`, which accurately collects memory metrics from container stats and host config. Additionally, it introduces a fetcher mock, aligning with the approach used for cgroups v1 and v2 fetchers, to facilitate integration testing. The PR also includes new integration tests that validate the memoryMetrics functionality.

**Note**: `SwapUsage` metric is missing since it's not reported the from Docker API.


src/raw/dockerapi/fetcher.go
```diff
@@ -35,24 +37,43 @@ func (f *Fetcher) Fetch(container types.ContainerJSON) (raw.Metrics, error) {
...
-	panic("unimplemented")
+func (f *Fetcher) memoryMetrics(containerStats types.StatsJSON, hostConfig *container.HostConfig) raw.Memory {
+	mem := raw.Memory{}
+
+	if containerStats.MemoryStats.Usage != 0 {
+		mem.UsageLimit = containerStats.MemoryStats.Limit
+		mem.FuzzUsage = containerStats.MemoryStats.Usage
+	}
+
+	if hostConfig != nil {
```
sigilioso, line 48: Could `hostConfig` be nil? If it could, should we log a debug message?

src/raw/dockerapi/fetcher.go
```diff
@@ -35,24 +37,43 @@ func (f *Fetcher) Fetch(container types.ContainerJSON) (raw.Metrics, error) {
...
+		// TODO: Disclaimer: we're inferring swap usage from provided metrics since Docker API doesn't expose it.
```
sigilioso, line 49: Is the TODO set to check later if the value matches the one fetched when reading values from the filesystem?

src/raw/dockerapi/fetcher.go
```diff
@@ -69,3 +90,41 @@ func (f *Fetcher) ContainerStats(ctx context.Context, containerID string) (types
 	}
 	return statsJSON, nil
 }
+
+// getUint64FromMapAndLog retrieves a uint64 value from a map based on the key, logs an error if the key is not found.
+func getUint64FromMapAndLog(m map[string]uint64, key string) uint64 {
+	if val, ok := m[key]; ok {
+		return val
+	}
+	log.Error("key %s not found in map", key)
```
sigilioso, line 99: Should we log an error? I wonder if a warning would be enough. I also wonder if it would be possible to add more information to the message. For instance, if ever encountered, `key file not found in map` wouldn't be very helpful, would it?

test/integration/dockerapi_fetcher_test.go
```diff
@@ -39,3 +45,72 @@ func logAsJSON(t *testing.T, title string, data any) {
...
+					MemoryReservation: 1024 * 1024 * 250,  // 250 MB, for SoftLimit calculation
+					MemorySwap:        1024 * 1024 * 1500, // Total memory + swap
+				},
+			},
+		},
+	}
+
+	// Fetch the metrics for the mocked container
+	fetcherMock := NewDockerAPIFetcherMock(mockContainerStats, nil)
+	metrics, err := fetcherMock.Fetch(mockedContainerJSON)
```
sigilioso, line 99: Did you consider defining this as an unit test instead? As I see it, we don't need to mock the fetcher (since we are calling the `Fetch` method directly) and we simply mock the underlying client.

It would be different from the approach taken for cgroups, but the cgroups use the fetcher through `biz.Processor` instead. If it useful, we could also add some integration tests but maybe we should use the the docker API directly. 

WDYT?

sigilioso (commented):
Looks good! It seems we can fetch all the metrics 🙂 

I've left some comments.

src/raw/dockerapi/fetcher.go
marcsanmi: Yes, it can be nil https://github.com/moby/moby/blob/master/api/types/types.go#L315C2-L315C39. Make sense, going to add a debug log in case of nil.

src/raw/dockerapi/fetcher.go
marcsanmi: As talked, we're going to remove this function and set it to 0.

src/raw/dockerapi/fetcher.go
marcsanmi: Agree, a warning might be better. However, what else would you add? I added the key since it's useful to see if what we are looking for... but the other part I see is to log the full stats map 🤔

sigilioso (approved):

src/raw/dockerapi/fetcher_test.go
```diff
@@ -69,3 +96,31 @@ func Test_NetworkMetrics(t *testing.T) {
...
+			Resources: container.Resources{
+				Memory:            1024 * 1024 * 500, // 500 MB
+				MemoryReservation: 1024 * 1024 * 250, // 250 MB, for SoftLimit calculation
+				MemorySwap:        1024 * 1024 * 600, // Total memory + swap
+			},
+		},
+	}})
+	require.NoError(t, err)
+
+	// Assert the fetched memory metrics match the expected outcome
```
gsanchezgavier, line 117: perhaps comparing the whole metrics against expecation struct?

src/raw/dockerapi/fetcher.go
```diff
@@ -35,8 +37,29 @@ func (f *Fetcher) Fetch(container types.ContainerJSON) (raw.Metrics, error) {
 	return metrics, nil
 }
 
-func (f *Fetcher) memoryMetrics(containerStats types.StatsJSON) raw.Memory {
-	return raw.Memory{}
+func (f *Fetcher) memoryMetrics(containerStats types.StatsJSON, hostConfig *container.HostConfig) raw.Memory {
+	mem := raw.Memory{}
+
+	if containerStats.MemoryStats.Usage != 0 {
```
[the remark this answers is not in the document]
sigilioso: The same is done when we fetch the values from cgroupsv2 files: https://github.com/newrelic/nri-docker/blob/1997998892d5d2e5697dd7b1221127d1440ebc19/src/raw/cgroupv2.go#L125
 I'm adding a comment explaining it.

src/raw/dockerapi/fetcher.go
```diff
@@ -35,8 +37,29 @@ func (f *Fetcher) Fetch(container types.ContainerJSON) (raw.Metrics, error) {
...
+		mem.UsageLimit = containerStats.MemoryStats.Limit
+		mem.FuzzUsage = containerStats.MemoryStats.Usage
```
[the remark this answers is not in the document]
sigilioso: I'd say it is OK because how it is got in cgroupsv2 file fetcher. However, let's be extra-careful when comparing these metrics values, just in case.

src/raw/dockerapi/fetcher.go
```diff
@@ -35,8 +37,29 @@ func (f *Fetcher) Fetch(container types.ContainerJSON) (raw.Metrics, error) {
...
+	}
+
+	if hostConfig != nil {
+		// Set SwapUsage to zero since it's not reported in the docker API. We're doing the same for Fargate.
+		mem.SwapUsage = 0
```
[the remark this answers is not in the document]
sigilioso: Right, the assignation is not needed, thanks!

src/raw/dockerapi/fetcher_test.go
```diff
@@ -50,6 +64,19 @@ var mockNetworkStats = types.NetworkStats{
 	TxDropped: mockNetworkValue,
 }
 
+var expectedMetrics = raw.Metrics{
```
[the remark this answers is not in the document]
sigilioso: changed: [7b66fd6](https://github.com/newrelic/nri-docker/pull/209/commits/7b66fd6c336bb96870db86d23b6db92050e00574)

src/raw/dockerapi/fetcher.go
```diff
@@ -82,3 +105,11 @@ func (f *Fetcher) containerStats(ctx context.Context, containerID string) (types
 	}
 	return statsJSON, nil
 }
+
+func getOrWarning(m map[string]uint64, key string, metricsPath string) uint64 { // nolint:unparam
+	if val, ok := m[key]; ok {
+		return val
+	}
+	log.Warn("Could not fetch metric value from docker API: the key %q was not found in %s", key, metricsPath)
```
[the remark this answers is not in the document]
sigilioso: done: [010adab](https://github.com/newrelic/nri-docker/pull/209/commits/010adab0447e77a9ac69154d7622310df7e6ca76)

src/raw/dockerapi/fetcher.go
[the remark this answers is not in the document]
sigilioso: I added the comment here: [010adab](https://github.com/newrelic/nri-docker/pull/209/commits/010adab0447e77a9ac69154d7622310df7e6ca76)

gsanchezgavier (approved):
thanks

PR #211, opened 2024-02-14
DockerAPI flag

paologallinaharbur:
Adding useDockerAP flag and marking old config as deprecated


src/config/config.go
```diff
@@ -10,6 +10,7 @@ type ArgumentList struct {
 	DockerClientVersion   string `default:"1.24" help:"Optional. Specify the version of the docker client. Used for compatibility."`
 	DisableStorageMetrics bool   `default:"false" help:"Disables storage driver metrics collection."`
 	ShowVersion           bool   `default:"false" help:"Print build information and exit"`
+	UseDockerAPI          bool   `default:"false" help:"Leverage Docker API to gather metrics instead of filesystem. This option is ignored if cgroupsV1 are detected"`
```
gsanchezgavier, line 13: just a nit but i would move this and the `Fargate` arg together and use a similar description , so is clear that the both flags change the source of the metrics. 

Ideally they should have a similar name like `UseFargate` and `UseDockerAPI` but that is not a nice breaking change.

src/docker.go
```diff
@@ -72,14 +73,30 @@ func populateFromDocker(i *integration.Integration, args config.ArgumentList) {
...
+		fetcher, err = raw.NewCgroupFetcher(args.HostRoot, cgroupInfo)
+		exitOnErr(err)
+	}
 
 	sampler, err := nri.NewSampler(fetcher, dockerClient, args)
 	exitOnErr(err)
 	exitOnErr(sampler.SampleAll(context.Background(), i, cgroupInfo))
 }
 
+func useDockerAPI(args config.ArgumentList, version string) bool {
```
gsanchezgavier, line 89: I would just pass the flag.
```suggestion
func useDockerAPI(useDockerAPI bool, version string) bool {
```

gsanchezgavier (approved):
thanks , just left some nits

src/docker.go
sigilioso: If we pass the flag only, let's use other name `userDockerAPI(useDockerAPI bool` is a bit odd, isn't it?

```go
func useDockerAPI(dockerAPIRequested bool, version string) bool {
```

maybe?

sigilioso (approved):

src/docker.go
paologallinaharbur: ok!

PR #213, opened 2024-02-14
Add network metrics test

gsanchezgavier:
refact integration test to check some of the generated network metrics.

Note:  I tried to generate "noise" for the rest of the metrics with `iperf` but the complexity was increasing.The following command could generate also `Txbytes` but i don't think is worth to add
```shell
apk add iperf
iperf -s & iperf -c localhost
```

Note:
The refact commit:
- moves around things
- replaces the implementation of `eventually` for the testify one


test/integration/dockerapi_fetcher_test.go
```diff
@@ -4,15 +4,16 @@ import (
...
 	"github.com/docker/docker/client"
 	"github.com/newrelic/nri-docker/src/raw/dockerapi"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
-// TODO: this should extended/replaced when the fetcher actually fetches something.
 // It currently it shows how to get container stats data and can be useful for development purposes.
-func TestDockerAPIHelpers(t *testing.T) {
+func TestDockerAPIFetcher(t *testing.T) {
```
sigilioso, line 16: Would it make sense to assure somehow we are executing the test in a system with cgroups-v2 to avoid false-positives?

I guess we won't face that issue in the CI. But, at least, I'd add a comment explaining the requirement.

test/integration/dockerapi_fetcher_test.go
```diff
@@ -28,10 +29,17 @@ func TestDockerAPIHelpers(t *testing.T) {
 	require.NoError(t, err)
 	logAsJSON(t, "Inspect data", &inspectData)
 
-	// fetch stats data
+	// Let stress container have some stresful moments before fetching data.
+	time.Sleep(time.Second * 10)
```
sigilioso, line 33: Did you consider using [testify's Eventually](https://pkg.go.dev/github.com/stretchr/testify/assert#Eventually), in the same way it is done "manually" in [other tests](https://github.com/newrelic/nri-docker/blob/4bd5599ca69d38f2a08c60e86bd4e8e72d2bd3a6/test/integration/metrics_test.go#L61)

sigilioso (commented):
Looks good!

I've left a couple of comments for us to discuss 🙂

test/integration/dockerapi_fetcher_test.go
gsanchezgavier: make sense!

gsanchezgavier: I added a check, as we mention if we need this check in other tests here it might be good to add a suite and checkit once

test/integration/dockerapi_fetcher_test.go
gsanchezgavier: I got rid of the "manual" implemention within the refactor.

.github/workflows/on_push_pr.yaml
```diff
@@ -15,7 +15,6 @@ jobs:
 
   test-integration-nix-cgroups-v1:
     name: Run integration tests on *Nix
-    needs: [push-pr]
     runs-on: ubuntu-20.04
```
sigilioso, line 18: It might be obvious and I think the renovate rules already ignore this, but we could we add a comment saying that requires an OS with cgroupsv1? (and we should not be updated). Just in case...

sigilioso (approved):
LGTM! 🔥

.github/workflows/on_push_pr.yaml
gsanchezgavier: there is a [step](https://github.com/newrelic/nri-docker/blob/e0f92c96aca758aca5e81abef616aa14c8eb83b9/.github/workflows/on_push_pr.yaml#L23) that check it :)

PR #214, opened 2024-02-14
NR-221185: add docker api cpu metrics

sigilioso:
Adds CPU metrics to the docker API fetcher.

Integration tests will be addressed in a different PR.


src/raw/dockerapi/fetcher.go
```diff
@@ -57,8 +58,27 @@ func (f *Fetcher) networkMetrics(containerStats types.StatsJSON) raw.Network {
 	return aggregatedMetrics
 }
 
-func (f *Fetcher) cpuMetrics(containerStats types.StatsJSON) raw.CPU {
-	return raw.CPU{}
+func (f *Fetcher) cpuMetrics(container types.ContainerJSON, containerStats types.StatsJSON) raw.CPU {
+	cpuStats := containerStats.CPUStats
```
gsanchezgavier, line 62: perhaps just passing this as func parameter?

gsanchezgavier (approved):
thanks just a nit

src/raw/dockerapi/fetcher.go
sigilioso: Changed here: [6c61dc8](https://github.com/newrelic/nri-docker/pull/214/commits/6c61dc89c32b737e0dbd8c3aec4ea60863883442)

PR #220, opened 2024-02-20
Test/compare metrics

paologallinaharbur:
This PR:
 - Introduces a simple test to compare metrics, some metrics are not compared since it does not make sense right now (they are either broken or missing)

### DIFF
--- a/test/integration/metrics_test.go
+++ b/test/integration/metrics_test.go
@@ -98,8 +98,8 @@ func TestCompareMetrics(t *testing.T) {
 			assert.InDelta(t, sampleCGroup.CPU.ThrottlePeriods, sampleAPI.CPU.ThrottlePeriods, 30, "ThrottlePeriods")
 			assert.InDelta(t, sampleCGroup.CPU.ThrottledTimeMS, sampleAPI.CPU.ThrottledTimeMS, 30, "ThrottledTimeMS")
 
-			assert.InDelta(t, sampleCGroup.BlkIO.TotalReadBytes, sampleAPI.BlkIO.TotalReadBytes, 20000000, "TotalReadBytes")
-			assert.InDelta(t, sampleCGroup.BlkIO.TotalWriteBytes, sampleAPI.BlkIO.TotalWriteBytes, 20000000, "TotalWriteBytes")
+			assert.InDelta(t, *sampleCGroup.BlkIO.TotalReadBytes, *sampleAPI.BlkIO.TotalReadBytes, 20000000, "TotalReadBytes")
+			assert.InDelta(t, *sampleCGroup.BlkIO.TotalWriteBytes, *sampleAPI.BlkIO.TotalWriteBytes, 20000000, "TotalWriteBytes")
 		},
 		eventuallyTimeout,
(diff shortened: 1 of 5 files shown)


sigilioso (approved):
Thanks!

I've left a couple of _nits_, feel free to ignore them 🙂

test/integration/metrics_test.go
```diff
@@ -11,15 +12,102 @@ import (
...
+
+			// TODO this comparisons should be enabled back as soon as they are fixed for cgroupsV2
+			// assert.InDelta(t, sampleCGroup.Memory.SoftLimitBytes, sampleAPI.Memory.SoftLimitBytes, 50000, "SoftLimitBytes")
+			// assert.InDelta(t, sampleCGroup.CPU.Shares, sampleAPI.CPU.Shares, 200, "Shares")
+
+			// These metrics cannot be tested since in the Github runner the metrics are not reported
+			// assert.InDelta(t, sampleCGroup.BlkIO.TotalWriteCount, sampleAPI.BlkIO.TotalWriteCount, 5000, "TotalWriteCount")
+			// assert.InDelta(t, sampleCGroup.BlkIO.TotalReadCount, sampleAPI.BlkIO.TotalReadCount, 500, "TotalReadCount")
+
+			// These metric is not available for fargate and through he DockerAPI
```
marcsanmi, line 82:
```suggestion
			// These metrics are not available for fargate and through the DockerAPI
```

I'd add the 2 @sigilioso [mentioned](https://github.com/newrelic/nri-docker/pull/220/files#r1502256593) in this section.

marcsanmi (approved):

src/raw/cgroupv2.go
```diff
@@ -119,6 +120,9 @@ func (cg *CgroupsV2Fetcher) cpu(metric *cgroupstatsV2.Metrics) (CPU, error) {
 
 func (cg *CgroupsV2Fetcher) memory(metric *cgroupstatsV2.Metrics) (Memory, error) {
 	mem := Memory{}
+	b, _ := json.Marshal(metric)
+	log.Error("This is the payload of memory metrics" + string(b))
```
gsanchezgavier, line 124: debugging leftover ?

test/integration/metrics_test.go
```diff
@@ -11,15 +12,102 @@ import (
...
+	metricsAPI := biz.NewProcessor(
+		persist.NewInMemoryStore(),
+		fetcherAPI,
+		dockerClient,
+		0)
+	sampleAPI, err := metricsAPI.Process(containerID)
+	require.NoError(t, err)
+
+	// Core metrics are calculated from metrics.Process time differences, using variables with seconds accuaracy. Use a tick larger than a second for accuracy.
+	time.Sleep(10 * time.Second)
```
gsanchezgavier, line 61: i think this is not needed since the "slow" ticker is already being used

gsanchezgavier (changes requested):
I think there is dangerous error log leftover

src/raw/cgroupv2.go
paologallinaharbur: Thanks for pointing it out, it was fixed locally with the other changes, but better to have a reminder! 😄

gsanchezgavier (approved):

PR #221, opened 2024-02-27
NR-231647: report only set blkio metrics, default to nil

marcsanmi:
This PR updates the logic to only report blkio metrics that are explicitly set, defaulting to nil for any metrics that are not.

### DIFF
--- a/src/biz/metrics.go
+++ b/src/biz/metrics.go
@@ -38,10 +38,10 @@ type Network raw.Network
 
 // BlkIO stands for Block I/O stats
 type BlkIO struct {
-	TotalReadCount  float64
-	TotalWriteCount float64
-	TotalReadBytes  float64
-	TotalWriteBytes float64
+	TotalReadCount  *float64
+	TotalWriteCount *float64
+	TotalReadBytes  *float64
+	TotalWriteBytes *float64
 }
 
 // CPU metrics
@@ -297,22 +297,36 @@ func (mc *MetricsFetcher) blkIO(blkio raw.Blkio) BlkIO {
 		if len(svc.Op) == 0 {
 			continue
 		}
+		var count = float64(svc.Value)
 		switch svc.Op[0] {
 		case 'r', 'R':
-			bio.TotalReadCount += float64(svc.Value)
+			if bio.TotalReadCount == nil {
+				bio.TotalReadCount = new(float64)
+			}
+			*bio.TotalReadCount += count
 		case 'w', 'W':
-			bio.TotalWriteCount += float64(svc.Value)
+			if bio.TotalWriteCount == nil {
+				bio.TotalWriteCount = new(float64)
+			}
+			*bio.TotalWriteCount += count
 		}
 	}
 	for _, bytes := range blkio.IoServiceBytesRecursive {
 		if len(bytes.Op) == 0 {
 			continue
 		}
+		var bCount = float64(bytes.Value)
 		switch bytes.Op[0] {
 		case 'r', 'R':
-			bio.TotalReadBytes += float64(bytes.Value)
+			if bio.TotalReadBytes == nil {
+				bio.TotalReadBytes = new(float64)
+			}
+			*bio.TotalReadBytes += bCount
 		case 'w', 'W':
-			bio.TotalWriteBytes += float64(bytes.Value)
+			if bio.TotalWriteBytes == nil {
+				bio.TotalWriteBytes = new(float64)
+			}
+			*bio.TotalWriteBytes += bCount
 		}
 	}
 	return bio
--- a/src/biz/metrics_aws_test.go
+++ b/src/biz/metrics_aws_test.go
@@ -43,10 +43,10 @@ func TestFargateMetrics(t *testing.T) {
 	assert.Equal(t, uint64(11), samples.Pids.Current)
 	assert.Equal(t, uint64(0), samples.Pids.Limit)
 
-	assert.Equal(t, float64(0), samples.BlkIO.TotalReadBytes)
-	assert.Equal(t, float64(0), samples.BlkIO.TotalReadCount)
-	assert.Equal(t, float64(7839744), samples.BlkIO.TotalWriteBytes)
-	assert.Equal(t, float64(957), samples.BlkIO.TotalWriteCount)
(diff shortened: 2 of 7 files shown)


test/integration/metrics_test.go
paologallinaharbur, line 1: can we add a new test for the logic just added?

src/nri/sampler.go
```diff
@@ -269,12 +269,25 @@ func pids(pids *biz.Pids) []entry {
...
+		if bio.TotalWriteBytes != nil {
+			sum += *bio.TotalWriteBytes
+		}
+		totalBytes = &sum
+	}
+
 	return []entry{
 		metricIOTotalReadCount(bio.TotalReadCount),
 		metricIOTotalWriteCount(bio.TotalWriteCount),
 		metricIOTotalReadBytes(bio.TotalReadBytes),
```
paologallinaharbur, line 288: shuoln't we check if bio.TotalReadBytes is nil?

paologallinaharbur: (same comment for all the functions)

paologallinaharbur: I fear we might introduce a nil pointer exception

src/biz/metrics_aws_test.go
```diff
@@ -118,3 +118,11 @@ func startMetadataEndpointStub(t *testing.T, taskID string) (server *httptest.Se
 	))
 	return ts, ts.Close
 }
+
+func assertBlkIOMetric(t *testing.T, expected float64, actual *float64) {
+	if actual != nil {
+		assert.Equal(t, expected, *actual)
+	} else {
+		assert.Nil(t, actual)
```
paologallinaharbur, line 126: would make sense depending on the test calling directly 

assert.Equal(t, expected, *actual)

or 
		assert.Nil(t, actual)
?

src/nri/sampler.go
marcsanmi: Updated.

src/biz/metrics_aws_test.go
marcsanmi: Done!

test/integration/metrics_test.go
marcsanmi: Added! 🙂

src/nri/sampler.go
```diff
@@ -269,17 +269,35 @@ func pids(pids *biz.Pids) []entry {
...
+	var totalBytes *float64
+	if bio.TotalReadBytes != nil || bio.TotalWriteBytes != nil {
+		sum := 0.0
+		if bio.TotalReadBytes != nil {
+			sum += *bio.TotalReadBytes
+		}
+		if bio.TotalWriteBytes != nil {
+			sum += *bio.TotalWriteBytes
+		}
+		totalBytes = &sum
```
gsanchezgavier, line 296: I don't think this is needed right?

gsanchezgavier: just `entries = append(entries, metricIOTotalBytes(sum))` and perhaps replacing the name of the varialble `sum` for `totalBytes`

gsanchezgavier (approved):
LGTM just left a nit

test/integration/metrics_test.go
```diff
@@ -280,6 +284,90 @@ func TestAllMetricsPresent(t *testing.T) {
...
+			client := mockDockerStatsClient{}
+			client.On("ContainerStats", mock.Anything).Return(tc.MockStats)
+			dockerAPIFetcher := dockerapi.NewFetcher(&client)
+
+			storer := inMemoryStorerWithPreviousCPUState()
+			inspector := NewInspectorMock(InspectorContainerID, InspectorPID, 2)
+
+			metrics := biz.NewProcessor(storer, dockerAPIFetcher, inspector, 0)
+
+			sample, err := metrics.Process(InspectorContainerID)
```
gsanchezgavier, line 364: couldn't the test use the SampleAll to cover more code and check the metrics in the sample? perhaps is not so hard at this point

src/nri/sampler.go
marcsanmi: Yes, we can simplify it by just using one scoped variable 👍

test/integration/metrics_test.go
marcsanmi: As discussed, this would require a refactor since `ContainerSampler` is living in the `nri` package and has all the fields unexported. So, we cannot directly use it in our integration tests. For more context, it's used as follows:
```golang
sampler := ContainerSampler{
    metrics: biz.NewProcessor(mStore, nil, mocker, 30*time.Minute),
    docker:  mocker,
    store:   mStore,
}
err = sampler.SampleAll(context.Background(), i, info)
```

src/nri/sampler.go
marcsanmi: Updated!

test/integration/common.go
```diff
@@ -83,3 +83,7 @@ func stress(t *testing.T, args ...string) (containerID string, closeFunc func())
 		}
 	}
 }
+
+func pointerToFloat64(f float64) *float64 {
```
sigilioso, line 87: Nit: `float65ToPointer` maybe?

src/biz/metrics.go
```diff
@@ -38,10 +38,10 @@ type Network raw.Network
...
 // BlkIO stands for Block I/O stats
 type BlkIO struct {
-	TotalReadCount  float64
-	TotalWriteCount float64
-	TotalReadBytes  float64
-	TotalWriteBytes float64
+	TotalReadCount  *float64
+	TotalWriteCount *float64
+	TotalReadBytes  *float64
+	TotalWriteBytes *float64
```
sigilioso, line 44: Since we don't control how values are encoded we cannot use json tags annotations instead of nil values, right?

Example:
```
TotalReadCount float64 `json:",omitempty"`
```

gsanchezgavier (approved):

src/biz/metrics.go
marcsanmi: Yep, they're different structs...

sigilioso: As discussed _offline_, it is not the same struct, so it is not possible 🤐

sigilioso (approved):
Thanks for taking care of this! 👍

PR #224, opened 2024-03-06
fix(swap): we should not report metrics if value is incorrect

paologallinaharbur:
This could be a breaking change. 
Old installations were reporting swap metrics even if no data was actually found for swap usage


src/biz/metrics.go
```diff
@@ -276,19 +249,62 @@ func (mc *MetricsFetcher) memory(mem raw.Memory) Memory {
...
+However, as the `docker stats` cli tool does, page cache is intentionally
+excluded to avoid misinterpretation of the output.
+Also mem.SwapUsage is parsed from memory.memsw.usage_in_bytes, which
+according to the documentation reports the sum of current memory usage
+plus swap space used by processes in the cgroup (in bytes). That's why
+Usage is subtracted from the Swap: to get the actual swap.
+*/
+func populateSwapUsageMetrics(mem raw.Memory, m Memory, swapLimit uint64) Memory {
+	var swapOnlyUsage uint64
+	if *mem.SwapUsage != 0 { // for systems that have swap disabled
```
gsanchezgavier, line 293: shouldnt check nil pointer?

src/biz/metrics.go
```diff
@@ -276,19 +249,62 @@ func (mc *MetricsFetcher) memory(mem raw.Memory) Memory {
...
+		swapOnlyUsage = *mem.SwapUsage - mem.FuzzUsage
 	}
+	swapUsage := mem.RSS + swapOnlyUsage
```
gsanchezgavier, line 296: so having `swapUsage` being reported even if `swapOnlyUsage` is always `0` (because `mem.SwapUsage`==nil) doesn't seem like an error to me  taking into account the [docs](https://docs.newrelic.com/attribute-dictionary/?event=ContainerSample&attribute=memorySwapUsageBytes) .
Probably the ideal would be not report the data since it will be the same as the mem.RSS but given that is already there perhaps we could keep just that and avoid avoid a "risky" breaking change

I think that

src/biz/metrics.go
marcsanmi: I think it's checked outside the function. However, maybe I'd like this to be self-contained in the function itself to avoid potential nil panic if we use that function somewhere else and we forget to check for `nil`.

paologallinaharbur: yes that function is called just in one place if mem.SwapUsage != nil. 

I can add the guard also inside, but it felt like a bit too much

src/biz/metrics.go
gsanchezgavier: for some reason I took for granted that there was no real swap memory, but what i proposed doesn't make sense if there is actually real swap memory.

src/biz/metrics.go
```diff
@@ -276,19 +249,62 @@ func (mc *MetricsFetcher) memory(mem raw.Memory) Memory {
...
+		RSSUsageBytes:    mem.RSS,
+		UsageBytes:       mem.RSS,
+		UsagePercent:     usagePercent,
+		KernelUsageBytes: mem.KernelMemoryUsage,
+		SoftLimitBytes:   softLimit,
+		SwapLimitBytes:   swapLimit,
+	}
+
+	if mem.SwapUsage != nil {
+		m = populateSwapUsageMetrics(mem, m, swapLimit)
```
gsanchezgavier, line 269: nit: I would create a sub-struct of metric that contains these metrics and return that from the func rather than mutate Memory. Or perhaps just have the logic here.

gsanchezgavier (approved):

PR #226, opened 2024-03-08
fix: io metrics collection

gsanchezgavier:
- Fixes io metrics collection
- Add sampler tests

### DIFF
--- a/src/biz/storage.go
+++ b/src/biz/storage.go
@@ -3,7 +3,7 @@ package biz
 import (
 	"fmt"
 
-	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/system"
 	humanize "github.com/dustin/go-humanize"
 )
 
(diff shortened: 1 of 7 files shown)


gsanchezgavier:
I will stop fixing legacy linter errors , the current ones are old ones that appear because of replacing the `system.Info` struct.

src/biz/storage.go
```diff
@@ -21,7 +21,7 @@ type DeviceMapperStats struct {
 
 // ParseDeviceMapperStats parses the output from the info DriverStatus info Array. This information is not consider stable by
 // Docker and could change anytime according to API the docs.
-func ParseDeviceMapperStats(info types.Info) (*DeviceMapperStats, error) {
+func ParseDeviceMapperStats(info system.Info) (*DeviceMapperStats, error) {
```
marcsanmi, line 24: Why this?

gsanchezgavier: the linter failed since the type is marked as deprecated

src/nri/sampler.go
```diff
@@ -274,16 +275,16 @@ func blkio(bio *biz.BlkIO) []entry {
 	var entries []entry
 
 	if bio.TotalReadCount != nil {
-		entries = append(entries, metricIOTotalReadCount(bio.TotalReadCount), metricIOReadCountPerSecond(bio.TotalReadCount))
+		entries = append(entries, metricIOTotalReadCount(*bio.TotalReadCount), metricIOReadCountPerSecond(*bio.TotalReadCount))
```
marcsanmi, line 278: Good catch! We're testing the biz metrics in the `TestBlkIOMetrics` but not the sampler 👍

marcsanmi (approved):
LGTM! 🚀

PR #227, opened 2024-03-08
chore: remove aws dep

gsanchezgavier:
- removes aws dep by copying a little bit of code
- bumps docker dep


paologallinaharbur (approved):
❤️

marcsanmi (approved):
Thanks for removing that noisy dep. The only thing I see is that we might need to check in case ARN format changes in the future 😁

PR #228, opened 2024-03-11
Prepare release

gsanchezgavier:
Adds changelog and merge from master commits

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,26 @@ Unreleased section should follow [Release Toolkit](https://github.com/newrelic/r
 
 ## Unreleased
 
+### bugfix
+- Updated logrus dependency to remediate vulnerability
+
+## v1.10.1 - 2024-03-05
+
+### ⛓️ Dependencies
+- Updated github.com/aws/aws-sdk-go to v1.50.31 - [Changelog 🔗](https://github.com/aws/aws-sdk-go/releases/tag/v1.50.31)
+
+## v1.10.0 - 2024-02-20
+
+### 🚀 Enhancements
+- Allow the path (TMP_DIR) to store container_cpus.json (cache file) to be modified by the agent
+
+### 🐞 Bug fixes
+- Metrics fetch is no longer executed for exited containers that was causing an error log line (`failed to open file: /proc/0/cgroup`) for each exited container. Exited containers samples are still being reported as they were, as far as the TTL (24hs by default) has not expired.
+
+### ⛓️ Dependencies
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,32 @@ Unreleased section should follow [Release Toolkit](https://github.com/newrelic/r
...
 ## Unreleased
 
+### enhancement
+- Add metrics collection from DockerAPI to support nodes with cgroup V2 on ECS clusters.
+
+### security
+- Updated logrus dependency to remediate vulnerability
+
+### breaking
+- Memory swap metrics `memorySwapUsageBytes`, `memorySwapOnlyUsageBytes` and `memorySwapLimitUsagePercent` will not be sent if not present. Before they were being reported with value `0` (#224)
```
marcsanmi, line 19: It's going to happen the same for the [blkio](https://github.com/newrelic/nri-docker/blob/NR-221180-docker-api-fetcher/src/nri/sampler.go#L274) metrics isn't it?

paologallinaharbur (approved):
Thanks for this scary PR 😄

CHANGELOG.md
paologallinaharbur, line 19: before they were reported with wrong values

gsanchezgavier: thanks i added the note

CHANGELOG.md
```diff
@@ -9,6 +9,35 @@ Unreleased section should follow [Release Toolkit](https://github.com/newrelic/r
 
 ## Unreleased
 
+### enhancement
+- Add metrics collection from DockerAPI to support nodes with cgroup V2 on ECS clusters.
+
+### bugfix
+- Blkio metrics `TotalReadCount`, `TotalWriteCount`, `TotalReadBytes` and `TotalWriteBytes` will not be sent if not present. Before they were being reported with value `0` (#221)
```
sigilioso, line 16: Should we move this to `breaking`? As I see it, it would be the same that the swap metrics bug-fix.

sigilioso (approved):
Thanks!

CHANGELOG.md
marcsanmi: I think there's a difference since swap metrics were being badly reported on Fargate, not with 0 values. Another thing is if we consider this a breaking change or not... we comment on it briefly [here](https://github.com/newrelic/nri-docker/pull/228#discussion_r1519803121). As said, I don't have a strong opinion in this case.

marcsanmi (approved):

PR #248, opened 2024-08-21
[NR-231640] fix: fetch SofLimitBytes from docker inspect

sigilioso:
`SoftLimitBytes` metric values was being fetched from `low` value in `memory.events` instead of `memory.low` (check [cgroups-v2 specs](https://git.kernel.org/pub/scm/linux/kernel/git/tj/cgroup.git/tree/Documentation/admin-guide/cgroup-v2.rst) for details).

Since the corresponding value is in the response of `docker inspect` (which is called in a previous step), this PR updates the cgroups-v2 fetcher to get it from there.

### DIFF
--- a/src/raw/cgroupv2.go
+++ b/src/raw/cgroupv2.go
@@ -41,11 +41,11 @@ func NewCgroupsV2Fetcher(
 
 // Fetch get the metrics that can be found in cgroups v2 file system
 // Unlike v1, cgroup v2 has only single hierarchy.
-func (cg *CgroupsV2Fetcher) Fetch(c types.ContainerJSON) (Metrics, error) {
+func (cg *CgroupsV2Fetcher) Fetch(containerInfo types.ContainerJSON) (Metrics, error) {
 	stats := Metrics{}
 
-	pid := c.State.Pid
-	containerID := c.ID
+	pid := containerInfo.State.Pid
+	containerID := containerInfo.ID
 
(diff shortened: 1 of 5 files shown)


sigilioso:
e2e are failing do to https://github.com/newrelic/newrelic-integration-e2e-action/pull/43

src/raw/cgroupv2.go
```diff
@@ -81,7 +81,7 @@ func (cg *CgroupsV2Fetcher) Fetch(c types.ContainerJSON) (Metrics, error) {
 		log.Error("couldn't get cpu count: %v", err)
 	}
 
-	if stats.Memory, err = cg.memory(metrics); err != nil {
+	if stats.Memory, err = cg.memory(metrics, c); err != nil {
```
paologallinaharbur, line 84: I like short names... but `c` is really short 😄 

(I know you inherited it, but if it is not complex I would change it to something more meaningful now that we are passing it down)

paologallinaharbur (approved):
Thanks for taking care!

src/raw/cgroupv2.go
sigilioso: That's right, let's improve it a little bit 🙂

sigilioso: Changed: [c91a728](https://github.com/newrelic/nri-docker/pull/248/commits/c91a728bfcf483f8639c96664f9eb3d05739af22)

paologallinaharbur (approved):

PR #250, opened 2024-08-22
Fix Swap metrics metrics

paologallinaharbur:
This PR:
 - makes sure there is no overflow. There is no change in the logic, I simply moved to an "early exit" approach to avoid nested ifs
 - removes a warning from the integration tests
 - reduce a test flackyness making the test less strict
 - fix an old integration test


src/biz/metrics.go
```diff
@@ -281,23 +281,35 @@ func (mc *MetricsFetcher) memory(mem raw.Memory) Memory {
...
-		m.SwapOnlyUsageBytes = &swapOnlyUsage
-		m.SwapLimitUsagePercent = &swapLimitUsagePercent
+	var swapOnlyUsage uint64
+	if *mem.SwapUsage != 0 { // for systems that have swap disabled
+		swapOnlyUsage = *mem.SwapUsage - mem.FuzzUsage
+	}
+	swapUsage := mem.RSS + swapOnlyUsage
+
+	swapLimitUsagePercent := float64(0)
+	// Notice that swapLimit could be 0 also if the container has swap enabled (=-1)
```
sigilioso, line 302:
```suggestion
	// Notice that swapLimit could be 0 also if the container swap has no limit (=-1)
```

test/integration/metrics_test.go
```diff
@@ -146,7 +146,9 @@ func TestHighCPU(t *testing.T) {
 			assert.True(t, cpu.KernelPercent >= 0,
 				"kernel percent not >= 0")
 
-			assert.Truef(t, cpu.UserPercent+cpu.KernelPercent <= cpu.CPUPercent,
+			// This test is flaky, the +2 should not be needed, but we noticed that from time to time due to race conditions,
+			// such value is slightly higher
+			assert.Truef(t, cpu.UserPercent+cpu.KernelPercent <= cpu.CPUPercent+1,
```
sigilioso, line 151: I think you mean `+1` in the comment. We could also use [assert.inDelta](https://pkg.go.dev/github.com/stretchr/testify/assert#InDelta)

sigilioso (approved):
Thanks for taking care of this!

test/integration/metrics_test.go
paologallinaharbur: Nice!

sigilioso (approved):
🚀

PR #256, opened 2024-09-20
Use storer from SDK (NR-302384)

kang-makes:
The implementation of the cache storer in this integration is the same as the one in the SDK at https://github.com/newrelic/infra-integrations-sdk/pull/305

CacheTTL removes to it uses the one from the SDK


src/nri/sampler.go
```diff
@@ -53,7 +48,7 @@ func NewSampler(fetcher raw.Fetcher, docker raw.DockerClient, config config.Argu
 	store, err := persist.NewFileStore(
 		persist.TmpPath(config.TempDir, "container_cpus"),
 		log.NewStdErr(true),
-		cacheTTL)
+		config.CacheTTL)
```
sigilioso, line 51: I guess we need to bump the sdk when https://github.com/newrelic/infra-integrations-sdk/pull/316 is released

kang-makes:
Superseded by #258

PR #264, opened 2024-11-19
use centralized builder image

rajrohanyadav:
Use the centralized docker image for building.

## How is it tested?
Created a pre-release and the action trigged successfully to build and push the artifacts to s3 in the correct location.

https://github.com/newrelic/nri-docker/actions/runs/11907011686

image


Makefile
```diff
@@ -1,12 +1,16 @@
...
+VALIDATE_DEPS    	= golang.org/x/lint/golint
+INTEGRATIONS_DIR 	= /var/db/newrelic-infra/newrelic-integrations/
+CONFIG_DIR       	= /etc/newrelic-infra/integrations.d
+GO_FILES        	:= ./src/
+GOOS             	= GOOS=linux
+GO               	= $(GOOS) go
+GO_VERSION 			?= $(shell grep '^go ' go.mod | awk '{print $$2}')
+UBUNTU_VERSION 		?= "16.04"
+BUILDER_IMAGE_TAG 	?= "latest"
+BUILDER_IMAGE 		?= "ghcr.io/newrelic/coreint-automation:latest-go$(GO_VERSION)-ubuntu$(UBUNTU_VERSION)"
```
paologallinaharbur, line 13: Oh I think I misunderstood how we were using it, so we leverage pre-compiled images, we do not trigger the creation each time

Makefile
```diff
@@ -1,12 +1,16 @@
...
+SRC_DIR          	= ./src/
+VALIDATE_DEPS    	= golang.org/x/lint/golint
+INTEGRATIONS_DIR 	= /var/db/newrelic-infra/newrelic-integrations/
+CONFIG_DIR       	= /etc/newrelic-infra/integrations.d
+GO_FILES        	:= ./src/
+GOOS             	= GOOS=linux
+GO               	= $(GOOS) go
+GO_VERSION 			?= $(shell grep '^go ' go.mod | awk '{print $$2}')
+UBUNTU_VERSION 		?= "16.04"
+BUILDER_IMAGE_TAG 	?= "latest"
```
paologallinaharbur, line 12: if no one is setting this I would not make it configurable and just place them in the BUILDER_IMAGE (that is already configurable)

Makefile
rajrohanyadav: Yes, we only trigger when we want a new image built for a new version, OR if some changes are pushed to main.

build/ci.mk
```diff
@@ -17,23 +18,23 @@ ci/debug-container: ci/deps
 			-e GPG_MAIL \
 			-e GPG_PASSPHRASE \
 			-e GPG_PRIVATE_KEY_BASE64 \
-			$(BUILDER_TAG) bash
+			$(BUILDER_IMAGE) bash
```
paologallinaharbur, line 21: why are we passing it here?

build/ci.mk
```diff
@@ -17,23 +18,23 @@ ci/debug-container: ci/deps
...
 
 .PHONY : ci/validate
 ci/validate: ci/deps
 	@docker run --rm -t \
 			--name "nri-$(INTEGRATION)-validate" \
 			-v $(CURDIR):/go/src/github.com/newrelic/nri-$(INTEGRATION) \
 			-w /go/src/github.com/newrelic/nri-$(INTEGRATION) \
-			$(BUILDER_TAG) make validate
+			$(BUILDER_IMAGE) make validate
```
paologallinaharbur, line 29: who is setting and then reading this environment variable? It comes from the makefile, and it is passed to a second makefile?

Makefile
rajrohanyadav: Makes sense. Done. Thanks for suggesting.

build/ci.mk
paologallinaharbur: ah ok I got it, it is replacing BUILDER_TAG

paologallinaharbur (approved):

alvarocabanas (approved):
LGTM, nice work

PR #271, opened 2025-02-04
feat(docker-win): add main for windows

rajrohanyadav:
* This modifies the module to enable adding windows functionality. 
* Certain decisions made to minimise duplication.
* Cgoups will only be supported for linux. Windows and darwin will use dockerAPI.

### DIFF
--- a/src/config/config.go
+++ b/src/config/config.go
@@ -6,7 +6,7 @@ type ArgumentList struct {
 	args.DefaultArgumentList
 	HostRoot              string `default:"" help:"If the integration is running from a container, the mounted folder pointing to the host root folder"`
 	Fargate               bool   `default:"false" help:"Enables fetching metrics from ECS Fargate. If enabled no metrics are collected from cgroups. Defaults to false"`
-	UseDockerAPI          bool   `default:"false" help:"Enables fetching metrics from Docker API. If enabled no metrics are collected from cgroups. This option is ignored if cgroupsV1 are detected"`
+	UseDockerAPI          bool   `default:"false" help:"Enables fetching metrics from Docker API. If enabled no metrics are collected from cgroups. This option is ignored if cgroupsV1 are detected or running on any OS other than linux"`
 	ExitedContainersTTL   string `default:"24h" help:"Enables to integration to stop reporting Exited containers that are older than the set TTL. Possible values are time-strings: 1s, 1m, 1h"`
 	DockerClientVersion   string `default:"1.24" help:"Optional. Specify the version of the docker client. Used for compatibility."`
 	DisableStorageMetrics bool   `default:"false" help:"Disables storage driver metrics collection."`
--- a/src/config/config.go
+++ b/src/config/config.go
@@ -6,7 +6,7 @@ type ArgumentList struct {
 	args.DefaultArgumentList
 	HostRoot              string `default:"" help:"If the integration is running from a container, the mounted folder pointing to the host root folder"`
 	Fargate               bool   `default:"false" help:"Enables fetching metrics from ECS Fargate. If enabled no metrics are collected from cgroups. Defaults to false"`
-	UseDockerAPI          bool   `default:"false" help:"Enables fetching metrics from Docker API. If enabled no metrics are collected from cgroups. This option is ignored if cgroupsV1 are detected or running on any OS other than linux"`
+	UseDockerAPI          bool   `default:"false" help:"Enables fetching metrics from Docker API. If enabled no metrics are collected from cgroups. For Linux: This option is ignored if cgroupsV1 are detected. For OSes other than Linux: This option is always ignored and defaults to true"`
 	ExitedContainersTTL   string `default:"24h" help:"Enables to integration to stop reporting Exited containers that are older than the set TTL. Possible values are time-strings: 1s, 1m, 1h"`
 	DockerClientVersion   string `default:"1.24" help:"Optional. Specify the version of the docker client. Used for compatibility."`
 	DisableStorageMetrics bool   `default:"false" help:"Disables storage driver metrics collection."`
--- a/src/docker.go
+++ b/src/docker.go
@@ -1,24 +1,15 @@
+// Copyright 2025 New Relic Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
 package main
 
 import (
-	"context"
-	"fmt"
 	"os"
-	"runtime"
 
-	"github.com/docker/docker/api/types/system"
-	"github.com/docker/docker/client"
 	"github.com/newrelic/infra-integrations-sdk/v3/integration"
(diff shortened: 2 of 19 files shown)


src/util/util.go
```diff
@@ -0,0 +1,55 @@
...
+
+	"github.com/docker/docker/api/types/system"
+	"github.com/newrelic/infra-integrations-sdk/v3/integration"
+	"github.com/newrelic/infra-integrations-sdk/v3/log"
+	"github.com/newrelic/nri-docker/src/config"
+	"github.com/newrelic/nri-docker/src/nri"
+	"github.com/newrelic/nri-docker/src/raw/aws"
+)
+
+const (
```
rahulreddy15, line 20: Should this be moved to `docker.go` ?

src/util/util.go
```diff
@@ -0,0 +1,55 @@
...
+	IntegrationName = "com.newrelic.docker"
```
rahulreddy15, line 21: Should this be moved to docker.go ?

src/util/util_linux.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2025 New Relic Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package util
```
rahulreddy15, line 4: Should there be a `//go:build linux` here ?
Should this be included non-linux cases ?

src/raw/cgroup.go
```diff
@@ -1,3 +1,5 @@
+//go:build linux
```
rahulreddy15, line 1: For windows and darwin we seem to be using the Docker API.
But, will building all of these c group files even for Windows and Darwin builds cause an issue ?

src/util/util_all.go
```diff
@@ -0,0 +1,35 @@
+// Copyright 2025 New Relic Corporation. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build !linux
```
rahulreddy15, line 4: Is using build flags like this a robust way of using the right file ?
Should `switch` or some other approach be used ?

src/util/util.go
rajrohanyadav: I can't think of a reason, why?

src/util/util_linux.go
rajrohanyadav: not needed, go automatically identifies based on file names.

src/raw/cgroup.go
rajrohanyadav: Yes, some system variables (from system libs) might not be available for these OS.

src/util/util_all.go
rajrohanyadav: IMO, build flags are the best way for this rather than sprinkling switch statements all throughout the code.

src/util/util_all.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/docker/docker/client"
+	"github.com/newrelic/infra-integrations-sdk/v3/integration"
+	"github.com/newrelic/nri-docker/src/config"
+	"github.com/newrelic/nri-docker/src/nri"
+	"github.com/newrelic/nri-docker/src/raw/dockerapi"
+)
+
+func UpdateDockerAPIArg(dockerAPIArg bool) bool {
+	return true
+}
```
sigilioso, line 21: Maybe we could improve the function name, the `Update` part suggest some modification and it is merely returning a value (even if it is used to modify the args right now)

src/docker.go
```diff
@@ -29,87 +20,23 @@ var (
 
 func main() {
 	args := config.ArgumentList{}
-	i, err := integration.New(integrationName, integrationVersion, integration.Args(&args))
-	exitOnErr(err)
+	// we always use the docker API for OSes other than Linux
+	args.UseDockerAPI = util.UpdateDockerAPIArg(args.UseDockerAPI)
```
sigilioso, line 24: We could also avoid modifying the `args` if possible. Maybe something like:

```go
useDockerAPI := util.shouldUseDockerAPI(args.UseDockerAPI)
```

We could also move the comment above to the OS specific implementation of the function.

sigilioso (commented):
Overall LGTM 👍  I've left some comments regarding minor things.

I guess more stuff needs to be changed as part a follow-up PRs: compiling for windows and linux, updating the readme, ....

rajrohanyadav:
> I guess more stuff needs to be changed as part a follow-up PRs: compiling for windows and linux, updating the readme, ....

Yes, more PRs will be there. Just wanted to help the reviewers with smaller PRs :)

src/config/config.go
```diff
@@ -6,7 +6,7 @@ type ArgumentList struct {
 	args.DefaultArgumentList
 	HostRoot              string `default:"" help:"If the integration is running from a container, the mounted folder pointing to the host root folder"`
 	Fargate               bool   `default:"false" help:"Enables fetching metrics from ECS Fargate. If enabled no metrics are collected from cgroups. Defaults to false"`
-	UseDockerAPI          bool   `default:"false" help:"Enables fetching metrics from Docker API. If enabled no metrics are collected from cgroups. This option is ignored if cgroupsV1 are detected"`
+	UseDockerAPI          bool   `default:"false" help:"Enables fetching metrics from Docker API. If enabled no metrics are collected from cgroups. This option is ignored if cgroupsV1 are detected or running on any OS other than linux"`
```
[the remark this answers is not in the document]
rajrohanyadav: Attempted to improve it in https://github.com/newrelic/nri-docker/pull/271/commits/19d53c83cac9e0a928185b97b9246c1bc9c7fd75
I believe we could default it to true for next releases, but not sure if that will be a breaking change 💭 
Let's keep it this way for this release just to de-couple things, and I'll try to see if it can be defaulted to true for all.

src/util/util_all.go
rajrohanyadav: Improved the function name in https://github.com/newrelic/nri-docker/pull/271/commits/023d6cc86ebc8781af89c2fed2235868974b6afa
Hmm, I didn't want someone to make a mistake of reading args somewhere in the program and then finding that it's value is different from the main caller. So, I updated the arg globally to avoid such mistakes. WDYT 🤔

src/config/config.go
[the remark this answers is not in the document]
sigilioso: 🤔 I'd say it would be a breaking change but I also think it is something to consider.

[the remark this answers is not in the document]
rajrohanyadav: Makes sense, I'll discuss with the PM and create a followup PR if it is agreed upon. Thanks.

src/util/util_all.go
sigilioso: I think it is OK, I usually prefer to keep the args values as they are and use different vars / functions or whatever is needed to bring the logic (build a different value from the args and use that value instead) but I don't have a strong oppinion 🙂

PR #274, opened 2025-02-12
feat(windows): add platform as a member to decide os specific metrics

rajrohanyadav:
There are no changes for Network metrics for windows. Therefore, just doing some housekeeping here. Adding platfrom to use for Memory and CPU metrics.


abhishuraina (approved):
LGTM

PR #276, opened 2025-02-25
feat(windows): add memory metrics for windows

rajrohanyadav:
```powershell
PS C:\Users\Administrator\Downloads> .\nri-docker.exe -pretty
[WARN] computing Storage Driver stats: only devicemapper is supported
{
        "name": "com.newrelic.docker",
        "protocol_version": "3",
        "integration_version": "0.0.0",
        "data": [
                {
                        "entity": {
                                "name": "af377adb75998abdba21b1369c51288958081365b07c9d6d808411abb174130c",
                                "type": "docker",
                                "id_attributes": []
                        },
                        "metrics": [
                                {
                                        "commandLine": "cmd",
                                        "containerId": "af377adb75998abdba21b1369c51288958081365b07c9d6d808411abb174130c",
                                        "cpuKernelPercent": 0,
                                        "cpuLimitCores": 4,
                                        "cpuPercent": 0,
                                        "cpuShares": 0,
                                        "cpuThrottlePeriods": 0,
                                        "cpuThrottleTimeMs": 0,
                                        "cpuUsedCores": 0,
                                        "cpuUsedCoresPercent": 0,
                                        "cpuUserPercent": 0,
                                        "event_type": "ContainerSample",
                                        "image": "sha256:89334c207742b3879d141bdc908f7335e7f4e5fdcfebd115a566e0e777661b43",
                                        "imageName": "mcr.microsoft.com/windows/nanoserver:ltsc2022",
                                        "memoryCommitBytes": 40091648,
                                        "memoryCommitPeakBytes": 108998656,
                                        "memoryPrivateWorkingSet": 30502912,
                                        "memoryUsageLimitPercent": 0.1809065050792046,
                                        "name": "laughing_clarke",
                                        "networkRxBytes": 506577,
                                        "networkRxBytesPerSecond": 0,
                                        "networkRxDropped": 0,
                                        "networkRxDroppedPerSecond": 0,
                                        "networkRxErrors": 0,
                                        "networkRxErrorsPerSecond": 0,
                                        "networkRxPackets": 295,
                                        "networkRxPacketsPerSecond": 0,
                                        "networkTxBytes": 39795,
                                        "networkTxBytesPerSecond": 0,
                                        "networkTxDropped": 0,
                                        "networkTxDroppedPerSecond": 0,
                                        "networkTxErrors": 0,
                                        "networkTxErrorsPerSecond": 0,
                                        "networkTxPackets": 347,
                                        "networkTxPacketsPerSecond": 0,
                                        "restartCount": 0,
                                        "shortContainerId": "af377adb7599",
                                        "state": "running",
                                        "status": "Up 2 hours",
                                        "threadCount": 0,
                                        "threadCountLimit": 0
                                }
                        ],
                        "inventory": {},
                        "events": []
                }
        ]
}
```

Memory Metrics for the windows containers:
```
"memoryCommitBytes": 40091648,
"memoryCommitPeakBytes": 108998656,
"memoryPrivateWorkingSet": 30502912,
"memoryUsageLimitPercent": 0.1809065050792046,
```


src/nri/sampler.go
```diff
@@ -257,6 +257,18 @@ func memory(mem *biz.Memory) []entry {
...
+	}
+	if mem.CommitPeakBytes > 0 {
+		metrics = append(metrics, metricMemoryCommitPeakBytes(mem.CommitPeakBytes))
+	}
+	if mem.PrivateWorkingSet > 0 {
+		metrics = append(metrics,
+			metricMemoryPrivateWorkingSet(mem.PrivateWorkingSet),
+			metricMemoryUsageLimitPercent(mem.UsagePercent),
+		)
+	}
```
sigilioso, line 271: Would it be possible to have values equal to zero here? We would omit the values.

If it was possible, we could use pointers instead and append the metrics only when values are not nil.

src/biz/metrics.go
```diff
@@ -265,6 +275,24 @@ func (mc *MetricsFetcher) memory(mem raw.Memory) Memory {
...
 	}
 
+	if mc.platform == constants.WindowsPlatformName {
+		m.CommitBytes = mem.Commit
+		m.CommitPeakBytes = mem.CommitPeak
+		m.PrivateWorkingSet = mem.PrivateWorkingSet
+
+		vmem, err := gops_mem.VirtualMemory()
+		if err != nil {
+			panic(err)
```
sigilioso, line 285: Is it Ok to panic here? Another possibility woudl be handling the error. Where the function is called:

Something similar to:
```go
metrics.Memory, err = memory(memory); if err != nil { /* */ }

File: src/nri/sampler_test.go [hunk 3]
```diff
@@ -234,6 +236,12 @@ func TestSampleAll(t *testing.T) {
 	assert.NotZero(t, metrics["memorySwapLimitUsagePercent"])
 	assert.NotZero(t, metrics["memorySoftLimitBytes"])
 
+	if runtime.GOOS == constants.WindowsPlatformName {
```
sigilioso, line 239: Would it be worth it to also check that those metrics are not there when `GOOS` is not windows?

src/biz/metrics.go
```diff
@@ -265,6 +275,24 @@ func (mc *MetricsFetcher) memory(mem raw.Memory) Memory {
...
+		}
+		totalMemory := vmem.Total
+		log.Debug("total memory on system: %v", totalMemory)
+		if totalMemory > 0 {
+			memoryUsagePercent := float64(m.PrivateWorkingSet) / float64(totalMemory) * 100
+			log.Debug("memory usage percent: %v", memoryUsagePercent)
+			m.UsagePercent = memoryUsagePercent
+		}
```
sigilioso, line 293: We are using the host's virtual memory to calculate the `UsagePercent`. Maybe future us would appreciate a comment explaining it.

sigilioso (commented):
Great job adding the windows memory metrics!

I've left some comments for us to discuss 🙂

src/nri/sampler.go
rajrohanyadav: Thanks for pointing it out, Good point, restructured the code to have different windows implementation. We will send the value that we get from dockerAPI. I don't feel a check if needed as we'll get the raw value from the dockerAPI. We send the value as 0 if we can't get it.

src/biz/metrics.go
rajrohanyadav: It is not at all Ok to panic here :D. I thought I'll fix this later and then forgot to fix it. Logging a warning now and sending 0 usage percentage.

src/biz/metrics.go
rajrohanyadav: Added a comment, I hope that is enough for future us (most of the times, it is not :D).

src/nri/sampler_test.go
rajrohanyadav: Restructured the code to not have these metrics for the other OS. So, I think it is not required to test this. But please let me know if I'm mistaken.

src/biz/metrics_all_test.go
```diff
@@ -0,0 +1,161 @@
...
+import (
+	"reflect"
+	"testing"
+	"time"
+
+	"github.com/newrelic/infra-integrations-sdk/v3/persist"
+	"github.com/newrelic/nri-docker/src/raw"
+)
+
+// swap tests are not applicable for windows
```
sigilioso, line 17: What about the specific metrics for windows? Do we need some sort of `metrics_windows_test` file?

src/biz/metrics_windows.go
```diff
@@ -0,0 +1,35 @@
...
+		CommitBytes:       mem.Commit,
+		CommitPeakBytes:   mem.CommitPeak,
+		PrivateWorkingSet: mem.PrivateWorkingSet,
+	}
+
+	vmem, err := gops_mem.VirtualMemory()
+	if err != nil {
+		log.Warn("error getting total memory on system: %v", err)
+	}
+	totalMemory := vmem.Total
```
sigilioso, line 20: When we get an error, is `vmem.Total` always 0?

src/nri/sampler_test.go
sigilioso: I think it is not that important (after all we are not testing that a metric name should not be there anywhere else) although it would be very explicit:
- For windows: I expect these metrics (and I also assure that the other metrics are not there).
- For the other OSs: I expect these metrics (and I also asure windows metrics are not set)

src/biz/metrics_windows.go
rajrohanyadav: Refactored in https://github.com/newrelic/nri-docker/pull/276/commits/c84e9aaced1086226c8f92894cfd73a282fdee25

ujwalsah (approved):
LGTM 🚀

PR #277, opened 2025-03-03
feat(windows): add cpu metrics for docker on windows

rajrohanyadav:
Add CPU metrics for docker on windows.
We rely on NumProcs from docker api to evaluate the total percentage of CPU used.
```json
{
        "name": "com.newrelic.docker",
        "protocol_version": "3",
        "integration_version": "0.0.0",
        "data": [
                {
                        "entity": {
                                "name": "2182a6f9073cebf88f3f578c6ad57803c6f4a4d78d4a471eb92caddf4fedc445",
                                "type": "docker",
                                "id_attributes": []
                        },
                        "metrics": [
                                {
                                        "commandLine": "cmd",
                                        "containerId": "2182a6f9073cebf88f3f578c6ad57803c6f4a4d78d4a471eb92caddf4fedc445",
                                        "cpuKernelPercent": 1.537019747728088,
                                        "cpuPercent": 0.384254936932022,
                                        "cpuProcs": 4,
                                        "cpuUserPercent": 1.537019747728088,
                                        "event_type": "ContainerSample",
                                        "image": "sha256:89334c207742b3879d141bdc908f7335e7f4e5fdcfebd115a566e0e777661b43",
                                        "imageName": "mcr.microsoft.com/windows/nanoserver:ltsc2022",
                                        "memoryCommitBytes": 40574976,
                                        "memoryCommitPeakBytes": 109862912,
                                        "memoryPrivateWorkingSet": 30769152,
                                        "memoryUsageLimitPercent": 0.18248551982744526,
                                        "name": "confident_rosalind",
                                        "networkRxBytes": 447204,
                                        "networkRxBytesPerSecond": 5.1647058823529415,
                                        "networkRxDropped": 0,
                                        "networkRxDroppedPerSecond": 0,
                                        "networkRxErrors": 0,
                                        "networkRxErrorsPerSecond": 0,
                                        "networkRxPackets": 231,
                                        "networkRxPacketsPerSecond": 0.058823529411764705,
                                        "networkTxBytes": 33186,
                                        "networkTxBytesPerSecond": 3.5647058823529414,
                                        "networkTxDropped": 0,
                                        "networkTxDroppedPerSecond": 0,
                                        "networkTxErrors": 0,
                                        "networkTxErrorsPerSecond": 0,
                                        "networkTxPackets": 272,
                                        "networkTxPacketsPerSecond": 0.058823529411764705,
                                        "restartCount": 0,
                                        "shortContainerId": "2182a6f9073c",
                                        "state": "running",
                                        "status": "Up About an hour",
                                        "threadCount": 0,
                                        "threadCountLimit": 0
                                }
                        ],
                        "inventory": {},
                        "events": []
                }
        ]
}
```


src/biz/metrics_all.go
```diff
@@ -99,3 +101,77 @@ func (mc *MetricsFetcher) memory(mem raw.Memory) Memory {
...
+	var previous struct {
+		Time int64
+		CPU  raw.CPU
+	}
+	// store current metrics to be the "previous" metrics in the next CPU sampling
+	defer func() {
+		previous.Time = metrics.Time.Unix()
+		previous.CPU = metrics.CPU
+		mc.store.Set(metrics.ContainerID, previous)
+	}()
```
sigilioso, line 115: I'm not sure if it is worth it because we are only repeating how the previous metrics are handled, but we could keep that in the common section. I guess it would be something like:

```go
// metrics.go
func (mc *MetricsFetcher) cpu(metrics raw.Metrics, json *types.ContainerJSON) CPU {
	var previous struct {
		Time int64
		CPU  raw.CPU
	}
	// store current metrics to be the "previous" metrics in the next CPU sampling
	defer func() {
		previous.Time = metrics.Time.Unix()
		previous.CPU = metrics.CPU
		mc.store.Set(metrics.ContainerID, previous)
	}()

	// Reading previous CPU stats
	if _, err := mc.store.Get(metrics.ContainerID, &previous); err != nil {
		log.Debug("could not retrieve previous CPU stats for container %v: %v", metrics.ContainerID, err.Error())
		return CPU{}
	}

    cpuMetrics(previous, metrics, json) // returns `CPU` and it has a different implementation for `_all` and `windows`
```

src/biz/metrics_windows.go
```diff
@@ -43,3 +46,69 @@ func (mc *MetricsFetcher) memory(mem raw.Memory) Memory {
...
+	passIntervals /= 100                                                     // Convert to number of 100ns intervals
+	passIntervals *= uint64(current.NumProcs)                                // Multiple by the number of processors
+
+	// Intervals used
+	intervalsUsed := current.TotalUsage - previous.TotalUsage
+
+	// Percentage avoiding divide-by-zero
+	if passIntervals > 0 {
+		log.Debug("passIntervals: %v, intervalsUsed: %v", passIntervals, intervalsUsed)
+		return float64(intervalsUsed) / float64(passIntervals) * 100.0
```
sigilioso, line 111: Nit: I'd swap the order to keep it consistent with the one in lines 91 and 94:

```go
return float64(intervalsUsed) *100.0  / float64(passIntervals)
```

src/biz/metrics_windows.go
```diff
@@ -43,3 +46,69 @@ func (mc *MetricsFetcher) memory(mem raw.Memory) Memory {
...
+	if durationIntervals <= 0 {
+		log.Debug("duration intervals is 0, returning empty CPU metrics")
+		return cpu
+	}
+
+	maxVal := float64(metrics.CPU.NumProcs * 100)
+
+	cpu.CPUPercent = cpuPercent(previous.CPU, metrics.CPU)
+
+	// If running in windows with Hyper-V isolation, can't check user and kernel CPU usage - set it to 0
```
sigilioso, line 83: Would it make sense to also cover this use-case in the windows unit-tests?

sigilioso (commented):
Looks great!

src/biz/metrics_all.go
rajrohanyadav: I tried this, but we should be able to report NumProcs or OnlineCPUs even if we don't have the last stored reading to calculate the percentage. If we move that to the new wrapper function, it would mean we return early and don't populate a few things we could. And if we leave that in the original function, we'd have to duplicate it. I'm feeling it wouldn't add much value though I'd love to explore other ways to refactor it.

src/biz/metrics_windows.go
rajrohanyadav: Makes sense, I changed the lines 91 and 94 to avoid potential overflows in https://github.com/newrelic/nri-docker/pull/277/commits/638e01c835eb28df5c75229fc722acbf548995d9

src/biz/metrics_windows.go
```diff
@@ -43,3 +46,69 @@ func (mc *MetricsFetcher) memory(mem raw.Memory) Memory {
 
 	return m
 }
+
+// TODO: use container inspect output to get the cpu limit
```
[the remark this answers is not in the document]
rajrohanyadav: Added the task to the todo. Yeah, we'll have to explore how to handle different config settings for CPU limit. Windows has three options, CPU Shares, CPU Count, and CPU Percentage with precedence defined. So, we need to ensure we have a good way to calculate the limit percentage accurately. That's why added it as a task. IMO, it would be better to address it once the integration with infra-agent is taken care of and we already have the basic functionality working for windows.

src/biz/metrics_windows.go
rajrohanyadav: Thanks for pointing it out. Yes, it should have been covered. Added in https://github.com/newrelic/nri-docker/pull/277/commits/1d838a5723c15c82b728f9955a2a67081a8a83e4

src/biz/metrics_windows.go
sigilioso: Thanks for also adding parentheses, it is clearer now 🙂

sigilioso (approved):
LGTM 🚀

ujwalsah (approved):
LGTM 🚀

PR #280, opened 2025-03-12
fix: use a separate config file for windows

rajrohanyadav:
Removed `file_exists: /var/run/docker.sock` from the linux config for windows.


rahulreddy15 (approved):
LGTM 🚀

PR #289, opened 2025-04-17
NR-403516: Move all pipelines from ubuntu-20 runners to ubuntu-latest

muruganishekar:
### Context

This pull request is driven by the upcoming deprecation of the Ubuntu 20.04 Actions runner image. GitHub has announced the following timeline for deprecation:

- **Deprecation Start Date**: 2025-02-01
- **Full Unsupported Date**: 2025-04-15

Given these dates, it is crucial to transition to supported runner images to ensure that our CI/CD pipelines do not encounter disruptions.

### Change Description

- **Objective**: Update all pipelines to use the `ubuntu-latest` runner image instead of the soon-to-be deprecated `ubuntu-20.xx`.
- **Background**: By updating to `ubuntu-latest`, we align our workflows with the latest stable and supported environment that GitHub provides. 
- **Impact**: Workflows using the `ubuntu-20.04` image label should be updated to `ubuntu-latest`.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -11,7 +11,7 @@ jobs:
   e2eTests-cgroups-v1:
     # Do not run e2e tests if commit message or PR has skip-e2e.
     if: ${{ ! contains(github.event.pull_request.labels.*.name, 'ci/skip-e2e') }}
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - name: Check cgroups version
         run: |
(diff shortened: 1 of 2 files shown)


rajrohanyadav:
Hi, I think we need to check why the checks failed and fix them.

rajrohanyadav:
> Hi, I think we need to check why the checks failed and fix them.

I see that default cgroup version starting Ubuntu 22 is v2 whereas it was v1 for Ubuntu 20. And we need to update GRUB bootloader settings if we want to use cgroups v1 which might not be possible with GH hosted runners. Therefore, I think it makes sense to remove these tests, but we should also document that we don't test and support croups v1 for `nri-docker` anymore.

rajrohanyadav (approved):
LGTM, should we consider removing tests themselves as we are not running them anymore.

